PR #309, opened 2023-02-16
Adding the rest of the modals to the earn borrow page

IanWoodard:
As the title suggests, I am adding the remainder of the modals to the earn borrow page and implementing the add collateral modal. The rest are waiting on some contracts to get deployed. Here is an image of the add collateral modal:
Screenshot 2023-02-16 at 9 20 25 AM

### DIFF
--- a/earn/src/components/borrow/ManageAccountButtons.tsx
+++ b/earn/src/components/borrow/ManageAccountButtons.tsx
@@ -2,13 +2,21 @@ import { OutlinedWhiteButtonWithIcon } from 'shared/lib/components/common/Button
 
 import { ReactComponent as PlusIcon } from '../../assets/svg/plus.svg';
 
-export default function ManageAccountButtons() {
+export type ManageAccountButtonsProps = {
+  onAddCollateral: () => void;
+  onRemoveCollateral: () => void;
+  onBorrow: () => void;
+  onRepay: () => void;
+};
+
+export default function ManageAccountButtons(props: ManageAccountButtonsProps) {
+  const { onAddCollateral, onRemoveCollateral, onBorrow, onRepay } = props;
   return (
     <div className='flex flex-col gap-2 w-max'>
       <OutlinedWhiteButtonWithIcon
         Icon={<PlusIcon />}
         position='leading'
-        onClick={() => {}}
+        onClick={onAddCollateral}
         size='S'
         svgColorType='stroke'
       >
@@ -17,7 +25,7 @@ export default function ManageAccountButtons() {
       <OutlinedWhiteButtonWithIcon
         Icon={<PlusIcon />}
         position='leading'
-        onClick={() => {}}
+        onClick={onRemoveCollateral}
         size='S'
         svgColorType='stroke'
       >
@@ -26,7 +34,7 @@ export default function ManageAccountButtons() {
       <OutlinedWhiteButtonWithIcon
         Icon={<PlusIcon />}
         position='leading'
-        onClick={() => {}}
+        onClick={onBorrow}
         size='S'
         svgColorType='stroke'
       >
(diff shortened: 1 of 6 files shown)


earn/src/components/borrow/ManageAccountButtons.tsx
```diff
@@ -2,13 +2,21 @@ import { OutlinedWhiteButtonWithIcon } from 'shared/lib/components/common/Button
...
 
-export default function ManageAccountButtons() {
+export type ManageAccountButtonsProps = {
+  onAddCollateral: () => void;
+  onRemoveCollateral: () => void;
+  onBorrow: () => void;
+  onRepay: () => void;
+};
+
+export default function ManageAccountButtons(props: ManageAccountButtonsProps) {
```
haydenshively, line 12: Since the modals are being defined in the parent component anyway, what's the point of defining the buttons here?

earn/src/components/borrow/modal/AddCollateralModal.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  const numericCollateralAmount = Number(collateralAmount) || 0;
+
+  const collateralAmountBig = useMemo(() => {
+    return new Big(numericCollateralAmount).mul(10 ** collateralToken.decimals);
+  }, [collateralToken.decimals, numericCollateralAmount]);
+
+  const existingCollateralBig = useMemo(() => {
+    return new Big(existingCollateral).mul(10 ** collateralToken.decimals);
+  }, [collateralToken.decimals, existingCollateral]);
```
haydenshively, line 184: Why `useMemo` for these? I don't see a reason for this simple calculation to be tied in with React's state management

earn/src/components/borrow/modal/AddCollateralModal.tsx
```diff
@@ -0,0 +1,275 @@
...
+          <Text size='XS' color={SECONDARY_COLOR} className='overflow-hidden text-ellipsis'>
+            You're adding{' '}
+            <strong>
+              {collateralAmount || '0.00'} {collateralToken.ticker}
+            </strong>{' '}
+            as collateral to the{' '}
+            <strong>
+              {marginAccount.token0.ticker}/{marginAccount.token1.ticker}
+            </strong>{' '}
+            pair. Your total collateral for this token in this pair will be{' '}
```
haydenshively, line 257: I think we should rephrase this a bit. Because they're not adding collateral to the _pair_, but rather to this particular smart wallet. For example, they may have 2 different smart wallets for the ETH/OP pair, in which case this is adding collateral to just 1 of their wallets, not both.

earn/src/components/borrow/modal/RemoveCollateralModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <Text size='XS' color={SECONDARY_COLOR} className='overflow-hidden text-ellipsis'>
+            You're removing{' '}
+            <strong>
+              {collateralAmount || '0.00'} {collateralToken.ticker}
+            </strong>{' '}
+            collateral from the{' '}
+            <strong>
+              {marginAccount.token0.ticker}/{marginAccount.token1.ticker}
+            </strong>{' '}
+            pair. Your total collateral for this token in this pair will be{' '}
```
haydenshively, line 113: Rephrase here as well probably.

earn/src/components/borrow/ManageAccountButtons.tsx
IanWoodard: To help reduce the size of the page component. Ideally, most of our components should be abstracted so that the page is concise. I could put the modals in here, but that would have its drawbacks, such as passing all of the data in here rather than just the listeners.

earn/src/components/borrow/modal/AddCollateralModal.tsx
IanWoodard: Good catch

haydenshively (approved):

PR #310, opened 2023-02-16
Starting to add multicall to optimize our requests

IanWoodard:
Multicall allows us to combine multiple calls that would otherwise be sent separately, which helps us reduce the number of outgoing requests we need to send to populate the data on the page. This PR switches some of our existing fetches to use multicall. There is still plenty more work to be done in future PRs.

### DIFF
--- a/earn/src/data/MarginAccount.ts
+++ b/earn/src/data/MarginAccount.ts
@@ -1,17 +1,18 @@
 import Big from 'big.js';
 import { secondsInYear } from 'date-fns';
-import { ethers } from 'ethers';
+import { ContractCallContext, ContractCallResults, Multicall } from 'ethereum-multicall';
+import { BigNumber, ethers } from 'ethers';
 import JSBI from 'jsbi';
 import { FeeTier, NumericFeeTierToEnum } from 'shared/lib/data/FeeTier';
 import { Address, Chain } from 'wagmi';
 
+import MarginAccountABI from '../assets/abis/MarginAccount.json';
+import MarginAccountLensABI from '../assets/abis/MarginAccountLens.json';
 import UniswapV3PoolABI from '../assets/abis/UniswapV3Pool.json';
 import VolatilityOracleABI from '../assets/abis/VolatilityOracle.json';
 import { makeEtherscanRequest } from '../util/Etherscan';
 import { toBig } from '../util/Numbers';
-// import { getAmountsForLiquidity, getValueOfLiquidity } from '../util/Uniswap';
-// import { UniswapPosition } from './actions/Actions';
-import { ALOE_II_FACTORY_ADDRESS, ALOE_II_ORACLE } from './constants/Addresses';
+import { ALOE_II_BORROWER_LENS_ADDRESS, ALOE_II_FACTORY_ADDRESS, ALOE_II_ORACLE } from './constants/Addresses';
 import { TOPIC0_CREATE_BORROWER_EVENT } from './constants/Signatures';
 import { BIGQ96 } from './constants/Values';
 import { Token } from './Token';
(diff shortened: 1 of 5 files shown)


earn/src/data/MarginAccount.ts
```diff
@@ -111,95 +119,147 @@ export type UniswapPoolInfo = {
...
+    });
+  });
+
+  const results = (await multicall.call(contractCallContext)).results;
+
+  let marginAccounts: MarginAccountPreview[] = [];
+
+  Object.values(results).forEach((value) => {
+    const contractResults = value.callsReturnContext;
+    const updatedReturnContext = convertBigNumbers(contractResults);
```
haydenshively, line 178: It seems like `convertBigNumbers` should operate on `contractResults.returnValues` instead of the entire `contractResults` object.

earn/src/data/MarginAccount.ts
```diff
@@ -111,95 +119,147 @@ export type UniswapPoolInfo = {
...
   const borrowAPR0 = interestRate0.eq('0') ? 0 : interestRate0.sub(1e12).div(1e12).toNumber() * secondsInYear;
-  const interestRate1 = new Big(lender1Basics.interestRate.toString());
+  const interestRate1 = new Big(lender1Basics[1].toString());
   const borrowAPR1 = interestRate1.eq('0') ? 0 : interestRate1.sub(1e12).div(1e12).toNumber() * secondsInYear;
-  const lender0Utilization = new Big(lender0Basics.utilization.toString()).div(10 ** 18).toNumber();
-  const lender1Utilization = new Big(lender1Basics.utilization.toString()).div(10 ** 18).toNumber();
-  const lender0TotalSupply = new Big(lender0Basics.totalSupply.toString());
-  const lender1TotalSupply = new Big(lender1Basics.totalSupply.toString());
-  const lender0TotalBorrows = new Big(lender0Basics.totalBorrows.toString());
-  const lender1TotalBorrows = new Big(lender1Basics.totalBorrows.toString());
```
haydenshively, line 202: Looking at the documentation, it seems like these named outputs should still exist. Why did you switch to the index-based stuff below?

earn/src/data/MarginAccount.ts
```diff
@@ -217,63 +277,84 @@ export async function fetchMarketInfoFor(
...
-  const assetsData = results[5];
-  const liabilitiesData = results[6];
   const [feeTier, oracleResult] = await Promise.all([
     uniswapPoolContract.fee(),
     volatilityOracleContract.consult(uniswapPool),
   ]);
 
   const assets: Assets = {
-    token0Raw: Big(assetsData.fixed0.toString())
+    token0Raw: Big(assetsData[0].toString())
```
haydenshively, line 333: Same question

earn/src/data/MarginAccount.ts
```diff
@@ -309,224 +390,6 @@ export function priceToSqrtRatio(price: number, token0Decimals: number, token1De
...
-
-//   // Constrain to be within TickMath's MIN_TICK and MAX_TICK
-//   if (a.lt('4295128740')) {
-//     a = new Big('4295128740');
-//   }
-//   if (b.gt('1461446703485210103287273052203988822378723970341')) {
-//     b = new Big('1461446703485210103287273052203988822378723970341');
-//   }
-//   return [a, b];
-// }
```
haydenshively, line 327: Is all this stuff being deleted because we don't need it for Earn?

IanWoodard: yes, I forgot to delete it earlier on.

earn/src/data/MarginAccount.ts
IanWoodard: Can you point me to where in the documentation you see this? I did not notice any named outputs for the results.

haydenshively: image

`returnValues` gives an object with an `amounts` property. That `amounts` label comes from the abi specified for the call.

IanWoodard: Maybe our abi is not getting passed in properly because I checked and there are no named outputs for the result when I do it.

earn/src/util/Multicall.ts
```diff
@@ -0,0 +1,14 @@
...
+  return callReturnContext.map((value) => {
+    value.returnValues = value.returnValues.map((returnValue) => {
+      if (returnValue?.type === 'BigNumber') {
+        returnValue = BigNumber.from(returnValue?.hex);
+      }
+      return returnValue as any[];
+    });
+    return value as CallReturnContext;
+  });
+}
```
haydenshively, line 14:
```suggestion
export function parseReturnValues(callReturnContext: CallReturnContext): any[] {
  return callReturnContext.returnValues.map((x) => x.type === 'BigNumber' ? BigNumber.from(x.hex) : x);
}
```

earn/src/data/MarginAccount.ts
```diff
@@ -111,95 +119,147 @@ export type UniswapPoolInfo = {
...
+
+  let marginAccounts: MarginAccountPreview[] = [];
+
+  Object.values(results).forEach((value) => {
+    const contractResults = value.callsReturnContext;
+    const updatedReturnContext = convertBigNumbers(contractResults);
+    const { feeTier, token0, token1, accountAddress, uniswapPool } = value.originalContractCallContext.context;
+    const assetsData = updatedReturnContext[0].returnValues;
+    const liabilitiesData = updatedReturnContext[1].returnValues;
+    const healthData = updatedReturnContext[2].returnValues;
```
haydenshively, line 182:
```suggestion
    const contractResults = value.callsReturnContext;
    const { feeTier, token0, token1, accountAddress, uniswapPool } = value.originalContractCallContext.context;
    const assetsData = parseReturnValues(contractResults[0]);
    const liabilitiesData = parseReturnValues(contractResults[1]);
    const healthData = parseReturnValues(contractResults[2]);
```

haydenshively (approved):

PR #311, opened 2023-02-17
Adding additional multicall usage

IanWoodard:
Given the success seen in the previous PR #310, I decided to move forward and add multicall to more places across the codebase. This change, in particular, helps us drastically cut down requests on the portfolio and markets page, causing a roughly 50% decrease in the number of outgoing requests.


earn/src/data/LendingPair.ts
```diff
@@ -77,79 +80,153 @@ export async function getAvailableLendingPairs(
...
+  const contractCallContexts: ContractCallContext[] = [];
+
+  addresses.forEach((market) => {
+    contractCallContexts.push({
+      reference: `basics0-${market.pool}`,
+      contractAddress: ALOE_II_KITTY_LENS_ADDRESS,
+      abi: KittyLensABI,
+      calls: [
+        {
+          reference: `basics0-${market.pool}`,
```
haydenshively, line 92: I don't think individual calls should have the same `reference` id as the outer contract. I guess if it works it works though.

earn/src/data/LendingPair.ts
```diff
@@ -77,79 +80,153 @@ export async function getAvailableLendingPairs(
...
-      const APY1 = utilization1 * (1 - 1 / 8) * (interestRate1.div(10 ** 12).toNumber() ** (365 * 24 * 60 * 60) - 1.0);
-
-      let IV = oracleResult[1].div(1e9).toNumber() / 1e9;
-      // Annualize it
-      IV *= Math.sqrt(365);
-
-      return new LendingPair(
+  const contractCallContexts: ContractCallContext[] = [];
+
+  addresses.forEach((market) => {
```
haydenshively, line 85: At some point this has got to break right? I can't imagine we can stuff 100 different lender requests into a single multicall, for example. Maybe worthwhile to paginate this somehow?

earn/src/data/LendingPair.ts
IanWoodard: I honestly threw this together somewhat hastily, so I will think about some other ways to do this.

earn/src/data/LendingPair.ts
IanWoodard: I just load-tested it and was able to do 6750 total calls without it failing. With that said, each lender does require more than one call, but that should still give us plenty of breathing room.

haydenshively (dismissed):

IanWoodard:
Revamped the logic to make it more precise and robust. Working towards the ideal way of doing multi-calls.

haydenshively (approved):

PR #312, opened 2023-02-17
Adding condition to stop unnecessary fetches

IanWoodard:
A long overdue update to the portfolio page stops the page from fetching data from within the modal unless that modal is open. This saves us a lot of requests as this not only happened on each initial render but also when users hovered over a different asset.


haydenshively (approved):
Good catch

PR #313, opened 2023-02-17
Adding caching for lender pair data

IanWoodard:
Building off of #312, this PR adds caching that prevents the unnecessary re-fetching of data for lending pairs. This also saves us a fair bit of outgoing requests and is a great optimization.

### DIFF
--- a/earn/src/components/portfolio/LendingPairPeerCard.tsx
+++ b/earn/src/components/portfolio/LendingPairPeerCard.tsx
@@ -142,6 +142,9 @@ export type LendingPairPeerCardProps = {
 export default function LendingPairPeerCard(props: LendingPairPeerCardProps) {
   const { activeAsset, lendingPairs } = props;
   const { activeChain } = useContext(ChainContext);
+
+  const [cachedData, setCachedData] = useState<Map<string, number>>(new Map());
+
   const options: DropdownOption<LendingPair>[] = useMemo(() => {
     return lendingPairs.map((lendingPair) => {
       return {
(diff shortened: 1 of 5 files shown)


earn/src/components/portfolio/LendingPairPeerCard.tsx
```diff
@@ -161,6 +164,11 @@ export default function LendingPairPeerCard(props: LendingPairPeerCardProps) {
...
   useEffect(() => {
     let mounted = true;
+    const cachedResult = cachedData.get(selectedOption.label);
+    if (cachedResult) {
+      setNumberOfUsers(cachedResult);
+      return;
+    }
     async function fetchNumberOfUsers() {
       const etherscanRequestLender0 = makeEtherscanRequest(
         7537163,
```
haydenshively, line 174: Unrelated, but this should be set to 0 or have different constants for different chains.

haydenshively (approved):
This looks good, but based on the PR description you gave, I'd expect `cachedData` to hold more than just the number of users. Are you still working on this?

earn/src/components/portfolio/LendingPairPeerCard.tsx
IanWoodard: Good point, will address in separate PR

IanWoodard:
> This looks good, but based on the PR description you gave, I'd expect `cachedData` to hold more than just the number of users. Are you still working on this?

What else is there to cache inside this component? I believe this is the only fetching that is done.

PR #314, opened 2023-02-17
Switching from block to zero to work cross chain

IanWoodard:
I am unsure if there are many downsides to not providing a reasonable fromBlock. I would like to know if it has any noticeable effect on the time it takes us to receive a response. If that is the case, we should use more precise numbers, which enables this to work cross-chain for now. We will likely not know for sure if there is much of any effect right now because we aren't fetching too much data, but as this amount grows, the impact may be more noticeable.


haydenshively (approved):
Yeah I doubt this'll impact the query time, but something to keep an eye on.

PR #316, opened 2023-02-18
Implementing the borrow modal on earn

IanWoodard:
Implementing the borrow modal on earn using the SimpleManager contract. If an ante is required, the user will see a message below the summary explaining what it is and why it is needed. I realize that, as of now, the ante is not returned to the user upon successfully repaying, but I know this is the goal. We can take that line out for now; it may even be gone by the time you read this. I also chose to forgo the logic determining the maximum amount a user can borrow. Because of this, if the user does try to borrow more than they can, I do not believe it will let them as the gas estimation would fail, but there is also no message sent to the user in that case, so this should still be addressed in a future PR. Additionally, the summary could use some work, but if it isn't too bad, we could off on changing it for now.  Below is an image of the updated component:
**No Ante:**
Screenshot 2023-02-18 at 5 53 46 PM
**Ante:**
Screenshot 2023-02-18 at 6 02 21 PM


earn/src/components/borrow/modal/BorrowModal.tsx
```diff
@@ -1,19 +1,183 @@
...
+      setIsPending(false);
+      setIsOpen(false);
+    } else if (!contractIsLoading && !contractDidSucceed) {
+      setIsPending(false);
+    }
+  }, [contractDidSucceed, contractData, contractIsLoading, setPendingTxn, setIsOpen]);
+
+  let confirmButtonState = ConfirmButtonState.READY;
+
+  if (false) {
```
haydenshively, line 114: ???

earn/src/components/borrow/modal/BorrowModal.tsx
```diff
@@ -31,7 +195,74 @@ export default function BorrowModal(props: BorrowModalProps) {
...
+            smart wallet. Your total liabilities for this token in this smart wallet will be{' '}
+            <strong>
+              {truncateDecimals(newLiability.toString(), borrowToken.decimals)} {borrowToken.ticker}
+            </strong>
+            .
+          </Text>
+          {shouldProvideAnte && (
+            <Text size='XS' color={TERTIARY_COLOR} className='overflow-hidden text-ellipsis'>
+              You will need to provide an additional {formattedAnte} ETH to cover the gas fees in the event that you are
+              liquidated. If you successfully repay your loan, you will be refunded the ETH.
```
haydenshively, line 242: Yeah we definitely need to take out the "you will be refunded" part.

earn/src/components/borrow/modal/BorrowModal.tsx
```diff
@@ -31,7 +195,74 @@ export default function BorrowModal(props: BorrowModalProps) {
...
+          <Text size='XS' color={SECONDARY_COLOR} className='overflow-hidden text-ellipsis'>
+            You're borrowing{' '}
+            <strong>
+              {borrowAmount || '0.00'} {borrowToken.ticker}
+            </strong>{' '}
+            using this{' '}
+            <strong>
+              {marginAccount.token0.ticker}/{marginAccount.token1.ticker}
+            </strong>{' '}
+            smart wallet. Your total liabilities for this token in this smart wallet will be{' '}
```
haydenshively, line 233: Instead of the textual summary here, it may be more useful to show a hypothetical health bar

haydenshively (approved):
Good for now, but we should definitely add a MAX button and a hypothetical health bar in a future PR

earn/src/components/borrow/modal/BorrowModal.tsx
IanWoodard: I mentioned this in the PR description; I didn't add logic for computing the maximum amount a user can borrow.

earn/src/components/borrow/modal/BorrowModal.tsx
IanWoodard: I'll see what I can do. With that said, I think it may be better to include both, so the user knows about both what is being transacted, and what the impact on their health is.

earn/src/components/borrow/modal/BorrowModal.tsx
haydenshively: Ok, if that's what this is intended to catch, I don't think "Unable to borrow XYZ" is the best button text. Maybe something like "Borrow amount exceeds limit" or "Account unhealthy"

IanWoodard: I wanted to leave this in as a reminder that the logic is not implemented. Also, because I didn't add the logic I didn't make a very specific title for the button. One of the reasons I didn't make it "account unhealthy" is because that doesn't factor in the case where there isn't enough supply.

haydenshively: I'm not sure I follow

IanWoodard: It doesn't matter what we set it to since that button state isn't being used in this PR. I feel that once we add in the logic to handle these states, we can add buttons for the corresponding states.

haydenshively: Then why have that enum value at all? If you're going to have it, the corresponding button text should be as correct as possible. If you _really_ don't want to do that here, then please make an issue about it so we don't forget.

IanWoodard: I can remove it

haydenshively (approved):

PR #317, opened 2023-02-18
Implementing the repay modal on earn

IanWoodard:
Builds off of #316. Implementing the repay modal on earn's borrow page. This includes the logic behind the max button and various states, such as when there is insufficient collateral or too much collateral (we could just repay the max in this case, but I wasn't sure what would be best here). Moreover, the summary includes the impact the action has on both the liabilities as well as collateral for the respective token. Below is an image of the updated component:
Screenshot 2023-02-18 at 5 54 43 PM


haydenshively (dismissed):
I realize now we should have a way for the user to deposit + repay simultaneously, instead of first having to deposit (via "Add Collateral") and then repaying in separate transactions. I'll make a new manager for this.

Fine to merge for now; I'll fiddle with things tonight

haydenshively (approved):

PR #321, opened 2023-02-20
Removing margin account previews from earn

IanWoodard:
As the title suggests, this PR aims to remove margin account previews from earn so that we can instead fetch all of the margin accounts simultaneously using a multicall. I also worked on improving the robustness and quality of our handling of the multicall data.


haydenshively (approved):
Yay!

PR #325, opened 2023-02-21
Fixing scientific notation breaking decimals

IanWoodard:
Fixing a bug that caused the repay modal to crash when the amount was small enough, and javascript decided to represent it in scientific notation in the toString result. To fix this, I added a function that removes scientific notation from a number (if a number is in scientific notation) and clamps that number to no more than a certain amount of decimals provided as input.

### DIFF
--- a/earn/src/components/borrow/modal/RepayModal.tsx
+++ b/earn/src/components/borrow/modal/RepayModal.tsx
@@ -1,6 +1,7 @@
 import { useContext, useState, useMemo, useEffect } from 'react';
 
 import { SendTransactionResult, FetchBalanceResult } from '@wagmi/core';
+import Big from 'big.js';
 import { ethers, BigNumber } from 'ethers';
 import { FilledStylizedButton } from 'shared/lib/components/common/Buttons';
 import { BaseMaxButton } from 'shared/lib/components/common/Input';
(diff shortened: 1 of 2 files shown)


earn/src/components/borrow/modal/RepayModal.tsx
```diff
@@ -136,7 +136,10 @@ function RepayButton(props: RepayButtonProps) {
 
   // MARK: Preparing data that's necessary to figure out button state -------------------------------------------------
   const existingLiability = marginAccount.liabilities[lender === marginAccount.lender0 ? 'amount0' : 'amount1'];
-  const bigExistingLiability = ethers.utils.parseUnits(existingLiability.toString(), repayToken.decimals);
+  const bigExistingLiability = ethers.utils.parseUnits(
+    removeScientificNotation(existingLiability.toString(), repayToken.decimals),
+    repayToken.decimals
+  );
```
haydenshively, line 142: I'm not sure this is the right solution. Instead, I think we should just do `ethers.utils.parseUnits(existingLiability.toFixed(repayToken.decimals), repayToken.decimals)`

haydenshively: Alternatively, `BigNumber.from(new Big(existingLiability).mul(10 ** repayToken.decimals))`

haydenshively (approved):

PR #329, opened 2023-02-22
Fixing bugs with add liquidity action card

IanWoodard:
Fixing bugs with switching between tokens in the add liquidity action card. Removing console logs as well.

### DIFF
--- a/prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
+++ b/prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
@@ -259,6 +259,13 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
     );
   }
 
+  let max0 = accountState.assets.token0Raw;
+  let max1 = accountState.assets.token1Raw;
+  // If token1 is selected, we need to swap the max amounts
+  if (!isToken0Selected) [max0, max1] = [max1, max0];
+  const maxString0 = Math.max(0, max0 - 1e-6).toFixed(6);
+  const maxString1 = Math.max(0, max1 - 1e-6).toFixed(6);
+
   const ticksAreDefined = previousLower != null && previousUpper != null && currentTick != null;
   const tickIncrement = (tickInfo && (isToken0Selected ? tickInfo.tickSpacing : -tickInfo.tickSpacing)) ?? null;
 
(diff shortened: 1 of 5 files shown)


prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
```diff
@@ -259,11 +259,6 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
     );
   }
 
-  const max0 = accountState.assets.token0Raw;
-  const max1 = accountState.assets.token1Raw;
-  const maxString0 = Math.max(0, max0 - 1e-6).toFixed(6);
-  const maxString1 = Math.max(0, max1 - 1e-6).toFixed(6);
-
```
haydenshively, line 266: For consistency with other action cards, can we keep the maxes up here? One option would be to do `let max0 = ...` and `let max1 = ...` and then you can simply do `if (!isToken0Selected) [max0, max1] = [max1, max0]` to flip them

prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
```diff
@@ -369,8 +364,14 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
...
   );
 
-  const input0MaxString = isToken0Selected ? maxString0 : maxString1;
-  const input1MaxString = isToken0Selected ? maxString1 : maxString0;
+  const maxes = [accountState.assets.token0Raw, accountState.assets.token1Raw];
+  const max0 = maxes[isToken0Selected ? 0 : 1];
+  const max1 = maxes[isToken0Selected ? 1 : 0];
+  const maxString0 = Math.max(0, max0 - 1e-6).toFixed(6);
+  const maxString1 = Math.max(0, max1 - 1e-6).toFixed(6);
+  // Not to be confused with token0Amount and token1Amount, these are the amounts associated with the inputs
```
haydenshively, line 372: I know what you mean but this comment is kinda confusing. Maybe just say something like "In the case that token0 is selected, we have to flip flop which amount is first and which is second"

haydenshively (approved):

prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
IanWoodard: I did some looking around, and it appears we are relatively inconsistent with this, but for this PR I agree we should stick with what it was. I plan on going through a second time in a separate PR and making these all consistent.

haydenshively (approved):

PR #333, opened 2023-02-23
Fetching uniswap positions

IanWoodard:
Added logic to fetch Uniswap positions efficiently using multi calls. We have yet to do anything with these positions, but this PR aims to set the foundation for what will come.


earn/src/data/Uniswap.ts
```diff
@@ -98,3 +121,111 @@ export function getValueOfLiquidity(position: UniswapPosition, currentTick: numb
...
+  const priceX96Big = new Big(priceX96.toString(10));
+
+  const decimalDiff = token0Decimals - token1Decimals;
+  const price0In1 = priceX96Big
+    .mul(10 ** decimalDiff)
+    .div(BIGQ96)
+    .toNumber();
+  const price1In0 = 1.0 / price0In1;
+  return isInTermsOfToken0 ? price0In1 : price1In0;
+}
```
haydenshively, line 153: I think we have functions like this in BalanceSheet already. Maybe we should create a file just for tick/sqrt/price conversions

haydenshively (approved):

haydenshively:
Also, you could go ahead and pass these in to the calculations for liquidation distance. I left TODOs so it should be clear where they're needed.

earn/src/data/Uniswap.ts
IanWoodard: In prime these are both in Uniswap.ts. While I am not against moving this stuff to a separate file it is out of scope of this PR

PR #341, opened 2023-02-25
Util function to avoid mistakes+overflows when fetching chain data

haydenshively:
Sometimes the health computed by the borrower lens was larger than Javascript could handle, resulting in Prime page crashing.

Also, the health on the earn Borrow page was not being divided by 1e18.

I added a new util function to make things clearer. Both issues are fixed now.

### DIFF
--- a/prime/src/data/MarginAccount.ts
+++ b/prime/src/data/MarginAccount.ts
@@ -6,7 +6,7 @@ import { Address, Chain } from 'wagmi';
 import UniswapV3PoolABI from '../assets/abis/UniswapV3Pool.json';
 import VolatilityOracleABI from '../assets/abis/VolatilityOracle.json';
 import { makeEtherscanRequest } from '../util/Etherscan';
-import { toBig } from '../util/Numbers';
+import { toBig, toImpreciseNumber } from '../util/Numbers';
 import { ALOE_II_FACTORY_ADDRESS, ALOE_II_ORACLE } from './constants/Addresses';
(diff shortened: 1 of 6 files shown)


prime/src/data/MarginAccount.ts
```diff
@@ -226,10 +202,10 @@ export async function fetchMarginAccount(
       assets: assets,
       liabilities: liabilities,
       sqrtPriceX96: toBig(oracleResult[0]),
-      health: health.div(1e9).toNumber() / 1e9,
+      health: health,
       lender0: lender0,
       lender1: lender1,
-      iv: oracleResult[1].div(1e9).toNumber() / 1e9,
+      iv: toImpreciseNumber(oracleResult[1], 18),
```
IanWoodard, line 208: Could we move this up with the health declaration to stay consistent?

IanWoodard (approved):

IanWoodard:
> Sometimes the health computed by the borrower lens was larger than Javascript could handle, resulting in Prime page crashing.
> 
> Also, the health on the earn Borrow page was not being divided by 1e18.
> 
> I added a new util function to make things clearer. Both issues are fixed now.

I thought health was at most 10,000.

haydenshively:
Health is hardcoded to be 10000 when liabilities are 0. But when liabilities are very very small (non-zero), it can be larger.

IanWoodard (approved):
LGTM

PR #342, opened 2023-02-25
Fix multiple MAX button bugs in Earn and Prime

haydenshively:
Builds on #334 

Fixes multiple bugs related to the MAX buttons
- on Prime's Borrow card, we were using the lender's initial available assets. If multiple cards are present and more than one modifies the available assets, that change wasn't being reflected in the MAX button math
- on Earn's Borrow modal, max was getting clamped to your amount of collateral. For example, if collateral is {1000 DAI, 0 USDC} and I try to borrow USDC, the max was being clamped to 0 even though it should be like 900.
- on both Earn and Prime, math for max withdraws was incorrect whenever the additional liabilities crossed the threshold from no shortfall --> shortfall. The big changes in `BalanceSheet` are to address this.


IanWoodard (approved):

IanWoodard (commented):
Really good work! Found a what may be a bug with the withdrawals:
Screenshot 2023-03-01 at 1 08 47 PM
Screenshot 2023-03-01 at 1 09 40 PM
Screenshot 2023-03-01 at 1 09 45 PM
Screenshot 2023-03-01 at 1 10 05 PM

haydenshively:
Good catch. As it turns out, I left out one of the divisors. Had it right on paper, wrong in code.

haydenshively:
@IanWoodard I'm going to go ahead and merge this. If we find additional bugs I can make a separate PR to address them.

PR #343, opened 2023-02-25
Adding the foundation for uniswap nfts

IanWoodard:
Laying the groundwork for our integrating with Uniswap NFTs by allowing users to deposit them as collateral. A few things to note:
- We fetch the uniswap nft positions on the borrow page since the fetching requires us to obtain all of the uniswap nft positions at once. This is fine due to our use of multicall meaning we only have to make a few calls to fetch many uniswap nft positions. Moreover, it enables us to filter the list of uniswap nft positions for the selected pool. 
- While I considered using the UniswapPosition type, it did not the granularity needed to filter by pair without wrapping it in more context. The good news about the new UniswapNFTPosition type is that it can easily be converted to a UniswapPosition; however, the UniswapPosition needs more context to be converted into a UniswapNFTPosition.
- I located and subsequently fixed a bug that was the result of passing the Token type through the multicall context, which stripped it of its instance methods. The fix involved passing the addresses and rebuilding the tokens from the other side.


earn/src/data/Token.ts
```diff
@@ -20,4 +20,12 @@ export class Token {
   get underlying(): Token {
     return this;
   }
+
+  equals(other: Token): boolean {
+    return (
+      this.chainId === other.chainId &&
+      this.address.toLowerCase() === other.address.toLowerCase() &&
+      this.decimals === other.decimals
```
haydenshively, line 28: We do not need to check `decimals`. Checking `chainId` and `address` guarantees equality.

earn/src/data/Uniswap.ts
```diff
@@ -18,6 +23,16 @@ export type UniswapPosition = {
 
 export type UniswapPositionPrior = Omit<UniswapPosition, 'liquidity'>;
 
+export type UniswapNFTPosition = {
+  operator: string;
+  token0: Token;
+  token1: Token;
+  fee: number;
+  tickLower: number;
+  tickUpper: number;
```
haydenshively, line 32: My personal convention is to just call these `lower` and `upper`

haydenshively (dismissed):
LGTM. Fetches look clean, nice work on that. Couple nits but don't wait for another review before merging

earn/src/data/Uniswap.ts
IanWoodard: I was following the naming convention from the contract.

PR #352, opened 2023-03-08
Adding arbitrum compatability

IanWoodard:
Adding new tokens and addresses to enable the Aloe II contracts to function on Arbitrum via the front end. Some of the token logos required a bit of manual finessing to get right, but other than that, most of the changes were either fix that make our logic more robust in the case that all of the data is not available or the tokens do not exist.


haydenshively (approved):
LGTM! Thanks for cleaning up the address namings.

PR #353, opened 2023-03-11
Beginning to add arbitrum to Prime

IanWoodard:
As the title suggests, I am adding arbitrum to Prime. This involved adding the necessary assets and fixing some bugs. There is still a bug involving the chain reconciliation on first load that I will add a fix for in a future PR.


prime/src/data/TokenData.ts
```diff
@@ -1,83 +1,24 @@
...
-);
-
-const WETH_OPTIMISM = new Token(
-  chain.optimism.id,
-  '0x4200000000000000000000000000000000000006',
-  18,
-  'WETH',
-  'Wrapped Ether',
-  WethLogo
-);
```
haydenshively, line 34: I prefer organizing them by chain rather than token, so that when we add new chains all the changes show up in one git diff block.

haydenshively (approved):

prime/src/data/TokenData.ts
IanWoodard: I just copy and pasted this from earn

PR #354, opened 2023-03-11
Improving initial chain loading logic

IanWoodard:
While this PR does not perfectly address the issue of initial chain loading but is an improvement, this is the best that can be done without a much more invasive approach.


prime/src/components/borrow/ManageAccountTransactionButton.tsx
```diff
@@ -197,11 +196,48 @@ export function ManageAccountTransactionButton(props: ManageAccountTransactionBu
...
+  const calldata = canConstructTransaction
+    ? ethers.utils.defaultAbiCoder.encode(['uint8[]', 'bytes[]', 'uint144'], [actionIds, actionArgs, positions])
+    : null;
+
+  const { config: contractConfig } = usePrepareContractWrite({
+    address: accountAddress,
+    abi: MarginAccountAbi,
+    functionName: 'modify',
+    chainId: activeChain.id,
+    args: [ALOE_II_FRONTEND_MANAGER_ADDRESS, calldata, [UINT256_MAX, UINT256_MAX]],
```
haydenshively, line 227: I believe the args at the end should be bools, not `UINT256_MAX`es. That should reduce gas slightly. Another thing we can do to reduce gas is to conditionally set them to true/false. The first arg only needs to be `true` when our list of actions contains:
- Removing token0 collateral
- Swapping token0 for token1

Similarly, the second arg only needs to be `true` when:
- Removing token1 collateral
- Swapping token1 for token0

Doing this will avoid unnecessary `approve`s in the contracts.

prime/src/components/borrow/ManageAccountTransactionButton.tsx
```diff
@@ -215,6 +225,44 @@ export function ManageAccountTransactionButton(props: ManageAccountTransactionBu
...
+  });
+  const isSwappingToken0ForToken1 = actionIds.some((id, idx) => {
+    if (id === 6) {
+      const actionArg: string = actionArgs[idx];
+      if (!actionArg) return false;
+      const firstArg: string = `0x${actionArg.slice(26, 66).toLowerCase()}`;
+      return firstArg === token0.address;
+    }
+    return false;
+  });
```
haydenshively, line 245: In theory you could combine the two `some` predicates and call the resulting boolean `shouldEnableTransfersForToken0`, but I think the verbosity is fine here.

haydenshively (approved):

prime/src/components/borrow/ManageAccountTransactionButton.tsx
IanWoodard: I saw this as well but felt being explicit was good; that way, it is clear what each piece is doing and what it means. I was thinking about abstracting the logic more but felt it wasn't worth it now.

PR #358, opened 2023-03-14
Starting to add multicall to prime

IanWoodard:
To start, I added three usages of mulitcall to cut the calls on the accounts page substantially down from ~50 to roughly ~4 or 5. I also added some usages to the actions page, but work still needs to be done.


prime/src/util/Multicall.ts
```diff
@@ -0,0 +1,19 @@
+import { CallReturnContext, ContractCallReturnContext } from 'ethereum-multicall';
+import { BigNumber } from 'ethers';
+
+export type ContractCallReturnContextEntries = {
+  [key: string]: ContractCallReturnContext;
+};
+
+export function convertBigNumbersForReturnContexts(callReturnContexts: CallReturnContext[]): CallReturnContext[] {
```
haydenshively, line 8: Could we put this in `shared`?

haydenshively (approved):
LGTM!

prime/src/util/Multicall.ts
IanWoodard: I'll do it in a future PR.

PR #359, opened 2023-03-14
Fixing some bugs found while adding arbitrum

IanWoodard:
Fixing a few bugs, such as a lack of a `+ 1` for the ante, a lack of a uniswap v3 subgraph URL for arbitrum (sadly the only one I found doesn't support our graphs, even uniswap doesn't have these graphs on arbitrum),  and cleaning up the liquidity chart preview to fit correctly at all sizes.

### DIFF
--- a/prime/src/App.tsx
+++ b/prime/src/App.tsx
@@ -47,6 +47,11 @@ export const theGraphUniswapV3Client = new ApolloClient({
   cache: new InMemoryCache(),
 });
 
+export const theGraphUniswapV3ArbitrumClient = new ApolloClient({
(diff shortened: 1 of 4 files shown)


haydenshively (approved):
That is unfortunate that the liquidity graphs don't work on Arbitrum.

Did the ANTE not work before? The contracts should require the +1, but if we're just noticing this now, it makes me think it was working without the +1?

IanWoodard:
> 

I neglected to test borrowing previously, and once I discovered it didn't work, I found the ANTE to be the culprit.

PR #360, opened 2023-03-15
Fixing etherscan on optimism

IanWoodard:
As the title suggests, Etherscan just made it so that we need separate API keys for optimism and mainnet leading us to update our code accordingly. Netlify should already have the new API keys in the env variables. Once merged, we just need to switch the normal Etherscan API key back (since I switched it to mitigate the issue initially).


haydenshively (approved):
We're definitely going to forget :(

IanWoodard:
> We're definitely going to forget :(

I won't 👍

PR #365, opened 2023-03-26
Fixing firefox bug for the second time since I forgot it existed

IanWoodard:
Fixing the same firefox bug from #294. I forgot about this and neglected to specify widths and heights for the SVGs causing errors on Firefox. This PR aims to fix that.


haydenshively (approved):
Nice catch lol

PR #370, opened 2023-04-06
Adding more health bars

IanWoodard:
Adding health bars to more borrow modals on earn. These are styled the same way as the previous health bars from the borrow modal.


haydenshively (commented):
Implementation looks good, but do you think we need this when adding collateral? I kinda think having for Borrow and Remove Collateral would be sufficient, since those are the actions that can _decrease_ health.

If we are going to have it for Add Collateral, we may as well add it to Repay too.

IanWoodard:
> good, but do you think we need this when adding collateral? I kinda think having for Borrow and Remove Collateral would be sufficient, since those are the actions that can _decrease_ health.
> 
> If we are going to have it for Add Collateral, we may as well add it to Repay too.

I am adding it to every modal that changes health. I am saving the repay modal for a separate PR.

haydenshively (approved):

PR #371, opened 2023-04-06
Adding interest rate to the earn borrow modal

IanWoodard:
Adding the hypothetical interest rate to the borrow modal so the user can see what the interest rate will be once they borrow. Below is an image of the updated component:
Screenshot 2023-04-04 at 11 21 49 AM


earn/src/data/MarketInfo.ts
```diff
@@ -0,0 +1,57 @@
...
+): Promise<MarketInfo> {
+  const [lender0Basics, lender1Basics] = await Promise.all([
+    lenderLensContract.readBasics(lender0),
+    lenderLensContract.readBasics(lender1),
+  ]);
+
+  const interestRate0 = toBig(lender0Basics.interestRate);
+  const borrowAPR0 = interestRate0.eq('0') ? 0 : interestRate0.sub(1e12).div(1e12).toNumber() * secondsInYear;
+  const interestRate1 = toBig(lender1Basics.interestRate);
+  const borrowAPR1 = interestRate1.eq('0') ? 0 : interestRate1.sub(1e12).div(1e12).toNumber() * secondsInYear;
```
haydenshively, line 36: I know this isn't your fault, but I think we forgot to multiply these borrow APRs by the (1 - 1/8) reserve factor. Check the APY calculation for reference.

earn/src/data/MarketInfo.ts
```diff
@@ -0,0 +1,57 @@
...
+  borrowerAPR1: number;
+  lender0Utilization: number;
+  lender1Utilization: number;
+  lender0TotalAssets: Big;
+  lender1TotalAssets: Big;
+  lender0TotalBorrows: Big;
+  lender1TotalBorrows: Big;
+  lender0AvailableAssets: Big;
+  lender1AvailableAssets: Big;
+};
```
haydenshively, line 21: Are the changes to this data structure meant to match Prime? Or is this entirely new?

IanWoodard: I just thought it would be better to give all the info since you removed some on prime that we use on here, so instead of having to recalculate the values using `lender0AvailableAssets` and `lender0TotalAssets`, I thought it would be easier also to include `lender0TotalBorrows`.

earn/src/data/MarketInfo.ts
IanWoodard: Is there anywhere else we need this?

haydenshively (approved):

IanWoodard:
@haydenshively Let me know if there is anywhere else I need to consider the reserve factor.

haydenshively (dismissed):

earn/src/data/MarketInfo.ts
```diff
@@ -0,0 +1,63 @@
...
+  ]);
+
+  const interestRate0 = toBig(lender0Basics.interestRate);
+  const borrowAPR0 = interestRate0.eq('0')
+    ? 0
+    : interestRate0.sub(1e12).div(1e12).toNumber() * secondsInYear * RESERVE_FACTOR;
+  const interestRate1 = toBig(lender1Basics.interestRate);
+  const borrowAPR1 = interestRate1.eq('0')
+    ? 0
+    : interestRate1.sub(1e12).div(1e12).toNumber() * secondsInYear * RESERVE_FACTOR;
```
haydenshively, line 42:
```suggestion
  const interestRate0 = toBig(lender0Basics.interestRate);
  const borrowAPR0 = interestRate0.eq('0')
    ? 0
    : interestRate0.sub(1e12).div(1e12).toNumber() * secondsInYear;
  const interestRate1 = toBig(lender1Basics.interestRate);
  const borrowAPR1 = interestRate1.eq('0')
    ? 0
    : interestRate1.sub(1e12).div(1e12).toNumber() * secondsInYear;
```

We don't multiple by `RESERVE_FACTOR` here because this is a borrow rate, not a supply rate.

earn/src/data/MarketInfo.ts
```diff
@@ -0,0 +1,63 @@
+import Big from 'big.js';
+import { secondsInYear } from 'date-fns';
+import { ethers } from 'ethers';
+import { Address } from 'wagmi';
+
+import { toBig, toImpreciseNumber } from '../util/Numbers';
+
+const RESERVE_FACTOR = 0.825; // (1 - 1/8)
```
haydenshively, line 8:
```suggestion
```

haydenshively (approved):

PR #372, opened 2023-04-08
Adding health bar to repay modal on earn

IanWoodard:
As the title suggests, I am adding the health bar to the repay modal to show the user what their health will be upon repaying to help users understand how to make their account healthier. The styling is the same as the other modals.


haydenshively (approved):
LGTM

PR #373, opened 2023-04-08
Adding error msg if liquidity chart doesn't have data

IanWoodard:
Currently, if there is no data for the liquidity chart, as is the case on Arbitrum, the chart will load indefinitely. This can lead to confusion among users as it is unclear what is happening. To address this, I made it so that the chart will disappear instead of loading forever if there is no chart data for a specific pool. Below is an image of the updated component:
Screenshot 2023-04-12 at 11 24 11 PM


haydenshively (commented):
I'm glad you explored this idea, but I think it would be better to just hide that section of the card entirely if data is unavailable. 2 reasons:
- saves space (and we've already gotten complaints that there's too much scrolling on Prime)
- imo it's pretty obvious that you can still add liquidity. if you couldn't, that entire card would be disabled. The simple fact that you're allowed to add it as an action and type numbers in implies that it's open for business. I don't think anyone will be confused by the lack of a graph.

haydenshively (approved):

PR #374, opened 2023-04-11
Interpolate to get evenly-spaced points on IV/CF graph

haydenshively:
Interpolates between data points (fetched from Etherscan) so that we get evenly spaced points on the graph for IV and CF

### DIFF
--- a/earn/src/components/borrow/BorrowGraph.tsx
+++ b/earn/src/components/borrow/BorrowGraph.tsx
@@ -14,12 +14,31 @@ import BorrowGraphTooltip from './BorrowGraphTooltip';
 const TEXT_COLOR = '#82a0b6';
 const GREEN_COLOR = '#82ca9d';
 const PURPLE_COLOR = '#8884d8';
-const MILLIS_PER_WEEK = 10 * 24 * 60 * 60 * 1000;
+const MILLIS_PER_DAY = 24 * 60 * 60 * 1000;
+const DAYS_TO_SHOW = 60;
+const NUM_DATA_POINTS = 100;
 
 export type BorrowGraphData = {
   IV: number;
   'Collateral Factor': number;
   x: Date;
+};
+
+const interp = (a: BorrowGraphData, b: BorrowGraphData, x: number): BorrowGraphData => {
+  const interpolated: BorrowGraphData = { x: new Date(x), IV: NaN, 'Collateral Factor': NaN };
+  const dx = b.x.getTime() - a.x.getTime();
+
+  if (x < a.x.getTime() - Number.EPSILON || x > b.x.getTime() + Number.EPSILON) {
+    throw new Error('extrap not interp');
+  }
+
+  const mIV = (b.IV - a.IV) / dx;
+  interpolated.IV = mIV * (x - a.x.getTime()) + a.IV;
+
+  const mCF = (b['Collateral Factor'] - a['Collateral Factor']) / dx;
+  interpolated['Collateral Factor'] = mCF * (x - a.x.getTime()) + a['Collateral Factor'];
+
+  return interpolated;
 };
 
 const Container = styled.div`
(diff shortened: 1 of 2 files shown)


earn/src/components/borrow/BorrowGraph.tsx
```diff
@@ -14,14 +14,33 @@ import BorrowGraphTooltip from './BorrowGraphTooltip';
...
   'Collateral Factor': number;
   x: Date;
 };
 
+const interp = (a: BorrowGraphData, b: BorrowGraphData, x: number): BorrowGraphData => {
+  const interpolated: BorrowGraphData = { x: new Date(x), IV: NaN, 'Collateral Factor': NaN };
+  const dx = b.x.getTime() - a.x.getTime();
+
+  if (x < a.x.getTime() - Number.EPSILON || x > b.x.getTime() + Number.EPSILON) {
+    throw new Error('extrap not interp');
```
IanWoodard, line 32: Remove?

earn/src/components/borrow/BorrowGraph.tsx
```diff
@@ -14,14 +14,33 @@ import BorrowGraphTooltip from './BorrowGraphTooltip';
...
+const DAYS_TO_SHOW = 60;
+const NUM_DATA_POINTS = 100;
 
 export type BorrowGraphData = {
   IV: number;
   'Collateral Factor': number;
   x: Date;
 };
 
+const interp = (a: BorrowGraphData, b: BorrowGraphData, x: number): BorrowGraphData => {
```
IanWoodard, line 27: 1. Could we change this from an arrow function to a regular function?
2. Could we call it `interpolate` rather than `interp`? 
3. Could we use better variable names?
4. Could we add jsDoc explaining the function and the arguments?

earn/src/components/borrow/BorrowGraph.tsx
```diff
@@ -90,11 +109,31 @@ export default function BorrowGraph(props: BorrowGraphProps) {
   const { graphData } = props;
 
   const now = Date.now();
-  const filteredGraphData = graphData
-    .filter((item) => now - item.x.getTime() <= MILLIS_PER_WEEK)
-    .map((item) => {
-      return { ...item, x: item.x.toISOString() };
-    });
+
```
IanWoodard, line 112: We should check that there is > 1 (since we already know it isn't empty); otherwise, we will run into an error.

earn/src/components/borrow/BorrowGraph.tsx
```diff
@@ -90,11 +109,31 @@ export default function BorrowGraph(props: BorrowGraphProps) {
...
+  const sortedGraphData = graphData.sort((item) => item.x.getTime());
+  const startIdx = sortedGraphData.findIndex((item) => now - item.x.getTime() <= MILLIS_PER_DAY * DAYS_TO_SHOW);
```
IanWoodard, line 114: I think we should also check that startIdx != -1 just to be safe. If we choose to not render in this case, we would need to move the useMediaQueryHook up above.

earn/src/components/borrow/BorrowGraph.tsx
```diff
@@ -90,11 +109,31 @@ export default function BorrowGraph(props: BorrowGraphProps) {
...
+
+  const t0 = sortedGraphData[startIdx].x.getTime();
+  const t1 = sortedGraphData[sortedGraphData.length - 1].x.getTime();
+  const dt = (t1 - t0) / NUM_DATA_POINTS;
+
+  const interpGraphData: BorrowGraphData[] = [];
+
+  let i = startIdx;
+  for (let t = t0; interpGraphData.length < NUM_DATA_POINTS - 1; t += dt) {
```
IanWoodard, line 123: Could we use a more conventional for loop structure?

IanWoodard (commented):
Left some initial comments,  looking good so far!

IanWoodard (approved):
One nit, but overall LGTM

earn/src/components/borrow/BorrowGraph.tsx
```diff
@@ -107,35 +110,39 @@ export type BorrowGraphProps = {
...
+
+  if (startIdx === -1 || startIdx === endIdx) return null;
 
+  // Timestamp of the first data point on the graph
   const t0 = sortedGraphData[startIdx].x.getTime();
-  const t1 = sortedGraphData[sortedGraphData.length - 1].x.getTime();
+  // Timestamp of the last data point on the graph
+  const t1 = sortedGraphData[endIdx].x.getTime();
+  // Time between consecutive points that's required to give us `NUM_DATA_POINTS` between `t0` and `t1`
   const dt = (t1 - t0) / NUM_DATA_POINTS;
```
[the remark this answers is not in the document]
haydenshively: I distinguished them on purpose. `dt` implies a small step as in calculus. That's what we're doing with the timesteps in the for loop. On the other hand, the change in X in the `interpolate` function is more like capital-D Delta (the change in a value over some longer time, as opposed to a tiny step).

If this were Swift I would've named it ΔX but sadly no can do here.

PR #375, opened 2023-04-12
Adding etherscan links to markets page

IanWoodard:
To make it easier for users to access the contracts, I added a more info button to each pair on the markets page that, when clicked, opens a modal with the symbol and address/etherscan link for each token/lender used for the pair. Below are images of the new component:
Screenshot 2023-04-12 at 12 42 46 PM
Screenshot 2023-04-12 at 12 42 35 PM


haydenshively (approved):
Looks great!

PR #377, opened 2023-04-14
Adding etherscan link to borrow page on earn

IanWoodard:
While I am not completely satisfied with how this looks, I think it is a good start and enables users to navigate to their smart wallet on Etherscan more easily. The link will use the correct block explorer for the active chain.
I added separate styling for mobile and desktop:
Screenshot 2023-04-14 at 9 59 54 AM
Screenshot 2023-04-14 at 10 00 34 AM

### DIFF
--- a/earn/src/data/MarginAccount.ts
+++ b/earn/src/data/MarginAccount.ts
@@ -31,7 +31,7 @@ export type Liabilities = {
  * For the use-cases that require all of the data
  */
 export type MarginAccount = {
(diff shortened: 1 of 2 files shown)


haydenshively (commented):
Pushed my suggestions [here](https://github.com/aloelabs/aloe-frontend/tree/hs-borrow-etherscan-link)

haydenshively (approved):

PR #381, opened 2023-04-15
Fixing numeric logic and adding proper gas estimation

IanWoodard:
Updating logic to utilize `Big`s whenever possible. Additionally, I am adding proper gas estimation.

### DIFF
--- a/earn/src/components/portfolio/modal/SendCryptoModal.tsx
+++ b/earn/src/components/portfolio/modal/SendCryptoModal.tsx
@@ -1,13 +1,13 @@
-import { useContext, useEffect, useState } from 'react';
+import { useContext, useEffect, useMemo, useState } from 'react';
 
 import { SendTransactionResult } from '@wagmi/core';
 import Big from 'big.js';
-import { BigNumber, ethers } from 'ethers';
+import { ethers } from 'ethers';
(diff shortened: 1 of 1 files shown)


earn/src/components/portfolio/modal/SendCryptoModal.tsx
```diff
@@ -200,6 +207,9 @@ export default function SendCryptoModal(props: SendCryptoModalProps) {
     setSelectedOption(defaultOption);
   }, [defaultOption]);
 
+  const bigSendAmount = new Big(sendAmountInputValue || 0).mul(String1E(selectedOption.decimals));
+  const bigSendBalance = new Big(depositBalance?.formatted ?? 0).mul(String1E(selectedOption.decimals));
```
haydenshively, line 211: Does `depositBalance` not have a `BigNumberish` property that we can convert from? I'd rather use that than the `formatted` thing, because we could avoid having to multiple by the `String1E` decimals value.

haydenshively (approved):
LGTM

earn/src/components/portfolio/modal/SendCryptoModal.tsx
IanWoodard: Acked, but holding off until we do the GoodNumber update.

PR #384, opened 2023-04-17
Add Permit2 hook and integrate in EarnInterestModal

haydenshively:


IanWoodard (approved):
I am noticing some errors in the logs relating to gas estimation failure whenever the amounts change after permitting, but the user is required to re-permit which is the correct behavior so I think we could look into this minor issue in a follow-up PR, but otherwise, LGTM.

PR #387, opened 2023-04-20
Move usePermit2 to shared lib

haydenshively:
I tried to move this out to its own repository so that others could use it, but ran into issues with React versioning and enum exports. This is the next best thing; moving to our shared lib lets us at least use it in Earn and Prime without copying code.


IanWoodard:
#392

PR #389, opened 2023-04-21
Automatically add earned fees into the P&L graph

IanWoodard:
As the title suggests, I moved the earned fees calculation out into the borrower action page so that we can use it to auto-populate the swap-fees input for the P&L graph. Please double-check my math and let me know if you think there are any places that may throw exceptions or cause any trouble. This will be a lot easier to be sure of once we start integrating GoodNumber into prime.


prime/src/pages/BorrowActionsPage.tsx
```diff
@@ -415,6 +425,61 @@ export default function BorrowActionsPage() {
...
+    const token1FeesEarnedInTermsOfToken0 = token0OverToken1.mul(token1FeesEarned).div(tokenDecimalDifference);
+    const token0FeesEarnedInTermsOfToken1 = token1OverToken0.mul(token0FeesEarned).div(tokenDecimalDifference);
+    const totalFeesEarned = isToken0Selected
+      ? token0FeesEarned.add(token1FeesEarnedInTermsOfToken0)
+      : token1FeesEarned.add(token0FeesEarnedInTermsOfToken1);
+    if (totalFeesEarned.gt(0)) {
+      const selectedTokenDecimals = isToken0Selected ? marginAccount.token0.decimals : marginAccount.token1.decimals;
+      setSwapFeesInputValue(totalFeesEarned.toFixed(selectedTokenDecimals));
+    }
+  }, [marginAccount, displayedMarginAccount, uniswapPositionEarnedFees, displayedUniswapPositions, isToken0Selected]);
```
haydenshively, line 481: While this math is correct, it's not consistent with how we would compute similar values on-chain. On-chain we compute everything in terms of token1 to avoid having denominators > numerators and entering the scary realm of decimals.

I'd like to do the same here -- just compute total earned fees in terms of token1, and only convert to token0 (if necessary) at the very end. Instead of what you currently have, where there are essentially 2 parallel paths of the math.

Another thing to address in a future PR would be to break this out into its own custom hook. Maybe we can do this when we switch to GoodNumber.

haydenshively (approved):
LGTM

prime/src/pages/BorrowActionsPage.tsx
IanWoodard: I just updated it, let me know if this is what you imagined.

prime/src/pages/BorrowActionsPage.tsx
```diff
@@ -415,6 +426,64 @@ export default function BorrowActionsPage() {
...
+    );
+    const token1FeesEarned = new Big(earnedFeesValues.reduce((p, c) => p.add(c.token1FeesEarned), new Big(0))).mul(
+      10 ** token1.decimals
+    );
+    const token0FeesEarnedInTermsOfToken1 = priceX96.mul(token0FeesEarned).div(BIGQ96);
+    const totalFeesEarnedInTermsOfToken1 = token1FeesEarned.add(token0FeesEarnedInTermsOfToken1);
+    if (totalFeesEarnedInTermsOfToken1.gt(0)) {
+      if (isToken0Selected) {
+        const totalFeesEarnedInTermsOfToken0 = totalFeesEarnedInTermsOfToken1.mul(BIGQ96).div(priceX96);
+        setSwapFeesInputValue(totalFeesEarnedInTermsOfToken0.div(10 ** token0.decimals).toFixed(token0.decimals));
```
haydenshively, line 480: To avoid trailing zeros, don't we omit the argument for `toFixed`?

prime/src/pages/BorrowActionsPage.tsx
```diff
@@ -415,6 +426,64 @@ export default function BorrowActionsPage() {
...
+      } else {
+        setSwapFeesInputValue(totalFeesEarnedInTermsOfToken1.div(10 ** token1.decimals).toFixed(token1.decimals));
```
haydenshively, line 482: Same here?

haydenshively (approved):
LGTM, thanks for going back through!

prime/src/pages/BorrowActionsPage.tsx
IanWoodard: I did this to avoid showing more decimals than the token allows (since it could technically have more).

prime/src/pages/BorrowActionsPage.tsx
IanWoodard: I did this to avoid showing more decimals than the token allows (since it could technically have more).

PR #392, opened 2023-04-24
Moving permit2 to shared

IanWoodard:
As the title suggests, I moved many files to shared to accompany the permit2. I also linted everything to get the code in order.


shared/src/abis/Permit2.ts
```diff
@@ -0,0 +1,901 @@
+export const Permit2ABI = [
```
haydenshively, line 1: Two things:
- Did you delete this from `earn/src`? git isn't recognizing it as a moved file.
- Let's stick with `permit2ABI` to match wagmi's capitalization style

shared/src/data/constants/Addresses.ts
```diff
@@ -0,0 +1 @@
+export const UNISWAP_PERMIT2_ADDRESS = '0x000000000022D473030F116dDEE9F6B43aC78BA3';
```
haydenshively, line 1: Now that we're using Typescript files for the ABIs, I think it'd be reasonable to put these sort of addresses in there. Especially when it's a universal address across all chains.

This would also make imports cleaner. e.g. `import {PERMIT2_ADDRESS, permit2ABI} from 'src/abis/Permit2.ts'`

haydenshively (dismissed):
Couple nits. Other than that just please double check that your merging of the two `Number.ts` files doesn't mess anything up layout-wise.

shared/src/abis/Permit2.ts
IanWoodard: Will delete, and I am cool with doing this.

PR #396, opened 2023-04-24
Adding GN to earn borrow modal

IanWoodard:
Starting to add GN to earn borrow modal, which involves updating some other data types such as MarketInfo, which are relatively simple to update compared to ones such as MarginAccount. There are some lingering usages of numbers and such that we should eliminate in the future, but for now, our goal is to add GN where we can.


shared/src/data/constants/ChainSpecific.tsx
```diff
@@ -10,3 +11,10 @@ export const CHAIN_LOGOS = {
   [chain.optimism.id]: <OptimismLogo width={16} height={16} />,
   [chain.arbitrum.id]: <ArbitrumLogo width={16} height={16} />,
 };
+
+export const ANTES = {
+  [chain.mainnet.id]: GN.fromDecimalString('0.001', 18),
+  [chain.goerli.id]: GN.fromDecimalString('0.001', 18),
+  [chain.optimism.id]: GN.fromDecimalString('0.001', 18),
+  [chain.arbitrum.id]: GN.fromDecimalString('0.001', 18),
+};
```
haydenshively, line 20: Noice

shared/src/abis/MarginAccount.ts
```diff
@@ -0,0 +1,404 @@
+export const marginAccountABI = [
```
haydenshively, line 1: Don't we already have `borrowerABI`?

shared/src/data/GoodNumber.ts
```diff
@@ -183,6 +183,14 @@ export class GN {
     return new GN(this.int.div(other).toFixed(0), this.decimals);
   }
 
+  recklessAdd(other: BigSource) {
+    other = new Big(other);
+    if (!isInteger(other)) {
+      console.warn(`recklessAdd by non-integer (${other.toString()}) wouldn't be possible in the EVM. Be careful!`);
+    }
+    return new GN(this.int.plus(other).toFixed(0), this.decimals);
+  }
```
haydenshively, line 192: This makes sense, but can we add `recklessSub` simultaneously?

haydenshively (approved):

haydenshively (approved):

PR #398, opened 2023-04-24
Fix APY calculation for lending

haydenshively:
If a full year passes between calls to `accrueInterest`, the previous calculation would be correct. If `accrueInterest` were being called every second, the new calculation would be correct. The true value is somewhere in-between, but the new one is smaller. We'd rather show users an APY that's too small than too large.


aneesh1 (approved):
$APY = (1 + \frac{r}{n})^n - 1$
$n = \text{Number of seconds in a year}$
Our $APR$ is already in units of percentage per second, so we have
$APY = (1 + APR)^{31536000} - 1$

Looks good!

IanWoodard (approved):
LGTM

PR #399, opened 2023-04-25
Cap utilization at 100% in computeYieldPerSecond

haydenshively:
Capped utilization at 100% in computeYieldPerSecond so that it doesn't give us scientific notation and break things.


IanWoodard (approved):
LGTM

PR #401, opened 2023-04-25
Add usePermit, useRedeem, and fix max redeem issues in withdraw modals

haydenshively:
`usePermit` is fairly straightforward and closely resembles `usePermit2`. `useRedeem` calls `usePermit` to give the Router permission to redeem the user's shares _**if**_ they're trying to redeem 95% or more of their shares **and** their `maxRedeem` value is changing. If redeeming less than 95% or `maxRedeem` is static, we use the standard `lender.redeem` function and do not need to permit anything.

Previously, the `WithdrawModal` was split into 2 large parts: the modal itself, and a `WithdrawButton`. I believe this was done for 2 reasons:
- it would've been a lot of logic to pack into one component
- conditional rendering (by returning `null` in `WithdrawModal`) would essentially make the `WithdrawButton` hooks conditional and only run when ready

However, we had duplicate calls to `convertToShares`. Since `useRedeem` now handles everything (including `convertToShares`), I decided to combine `WithdrawModal` and `WithdrawButton` into a single simplified component.

Events when entering an amount, _before this change:_
- 1 modal render (sometimes 2, unclear why)
- 4 button renders (sometimes 8, unclear why)
- 2 network requests (sometimes 4, unclear why)

Events when entering an amount, _after this change:_
- 5 modal renders
- 2 network requests (when redeeming normally) (one is `convertToShares`, one is gas estimation)
- 4 network requests (when redeeming via permit with router) (`convertToShares`, permit data fetches, and gas estimation)

I spent a very long time trying to optimize the number of renders and requests, but no matter how many useMemo's I used, I got the same results. Hypothetically if we abandon some of the wagmi hooks in favor of vanilla ethers, we could cut down on renders, but I don't think that's worth it. Similarly, we could cut out network requests if we used recklessly unprepared args for `useContractWrite`.

I don't think we should worry about any of that, but I did leave a TODO in there to consider debouncing the text input.


earn/src/data/hooks/UseRedeem.ts
```diff
@@ -0,0 +1,167 @@
...
+    write: redeem,
+    data: redeemTxn,
+    isLoading: isAskingUserToRedeem,
+    reset: resetRedeemTxn,
+  } = useContractWrite({
+    ...configRedeem,
+    request: configRedeem.request
+      ? {
+          ...configRedeem.request,
+          gasLimit: configRedeem.request?.gasLimit.mul(110).div(100),
```
IanWoodard, line 97: Could we make the 110 into a constant with the same name we usually use? I plan to replace those with a universal constant, and making it a constant would make it easier for me to find it.

earn/src/components/portfolio/modal/WithdrawModal.tsx
```diff
@@ -7,254 +7,136 @@ import { BaseMaxButton } from 'shared/lib/components/common/Input';
...
-    <FilledStylizedButton
-      size='M'
-      onClick={() => handleClickConfirm()}
-      fillWidth={true}
-      disabled={shouldConfirmButtonBeDisabled}
-    >
-      {confirmButton.text}
-    </FilledStylizedButton>
-  );
+function doesContainToken(token: Token, pair: LendingPair) {
```
IanWoodard, line 64: I feel this should be called something like `doesLendingPairContainToken`, and the first arg should be the pair and the second the token.

IanWoodard (approved):
After chatting about this offline, I think we got many bugs fixed. A few nits, but otherwise LGTM

IanWoodard (approved):
LGTM

PR #403, opened 2023-04-28
Adding unit tests for GN

IanWoodard:
Added some initial unit tests for GN, which will hopefully lay the groundwork for future unit tests across the repo. I opted to use jest with ts-jest. Currently, tests are only found in shared, but that will likely be different in the future as we expand our unit testing efforts.


shared/src/data/GoodNumber.test.ts
```diff
@@ -0,0 +1,349 @@
...
+
+  describe('zero', () => {
+    it('should return a zero value', () => {
+      expect(GN.zero(18).toString(GNFormat.DECIMAL)).toEqual('0');
+    });
+  });
+
+  describe('fromDecimalBig', () => {
+    it('should return a decimal string', () => {
+      expect(GN.fromDecimalBig(new Big('1'), 18).toString(GNFormat.DECIMAL)).toEqual('1');
```
haydenshively, line 313: I'd like to see more cases here. For example, what happens if the BigSource is `'0.1'`?

haydenshively (approved):
Seems legit. I'd like to see a few more test cases for the factories, but we can add those later.

shared/src/data/GoodNumber.test.ts
IanWoodard: Will add follow-up PR(s) with more tests. This is just to get us started.

PR #404, opened 2023-04-28
[Earn] Adjusting icon styling in the asset bar 

IanWoodard:
Updated the styling of the icon in the asset bar to ensure they don't blend into the background. This is accomplished by adding a white background to help tokens with a transparent icon stand out against their prominent colored background. A thin black border surrounds the icons as it creates a bit of separation and is the norm throughout the repo.
image
image
image
image


earn/src/components/portfolio/AssetBar.tsx
```diff
@@ -50,8 +50,9 @@ const AssetChunkContainer = styled.div.attrs((props: { percentage: number; color
 const AssetIcon = styled.img`
   width: 32px;
   height: 32px;
-  border: 2px solid transparent;
+  border: 1px solid #000000;
```
haydenshively, line 53: Can we make the border match our site's background color, instead of black?

haydenshively (approved):

PR #405, opened 2023-04-28
Sort lending pairs by user's balance in withdraw dropdown

haydenshively:


IanWoodard (approved):
LGTM

IanWoodard (changes requested):
Check if lendingPairs and lendingPairBalances can have different lengths. Maybe add an additional check?

PR #406, opened 2023-04-28
Fixing max withdraw not letting user click

IanWoodard:
Recently, we introduced logic to disable the `convertToShares` fetch if the amount input is greater than the user's balance. While this logic is correct, it didn't take the max button's implementation into account. When the user clicks the max button, we pass Q112 as the amount to indicate the user wants to completely drain the account. However, Q112 is greater than the user's balance causing the `convertToShares` fetch to remain disabled and stopping the entire flow in its tracks. To fix this, I added an additional check for whether the amount is equal to Q112, in which case we enable the `convertToShares` fetch. While this works, I do think we could use a constant that is used by both the logic that sets the max value to Q112 and the check to see if the amount is equal to it so that we don't accidentally only change one of them at some point in the future.

### DIFF
--- a/earn/src/data/hooks/UseRedeem.ts
+++ b/earn/src/data/hooks/UseRedeem.ts
@@ -63,11 +63,13 @@ export function useRedeem(chainId: number, lender: Address, amount: GN, owner: A
   });
   const [maxAmount, maxShares, maxSharesIsChanging] = maxData ?? [BN0, BN0, false];
 
+  // If the user is trying to redeem more than they can, we'll just redeem the max.
+  // This means we won't refetch multiple times if the user enters a number greater than the max.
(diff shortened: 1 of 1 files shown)


haydenshively (commented):
Good find! As you alluded to, this fix depends on both the hook and the modal using Q112 to represent the max value. To eliminate that dependency while fixing the bug and preserving our "don't refetch when user types 9999999999" behavior, I suggest the following:

```js
const amountToConvert = amount.toBigNumber().lte(maxAmount) ? amount : GN.Q(112);
const { data: sharesData, isFetching: isFetchingShares } = useContractRead({
  ...erc4626,
  functionName: 'convertToShares',
  args: [amountToConvert.toBigNumber()],
});
```

haydenshively (approved):

PR #408, opened 2023-04-29
Updating sentry, splitting it into two projects

IanWoodard:
To provide better error visibility, I updated Sentry and split it into two projects, one for Earn and one for Prime. This allows us to track them separately, which I see as a win, but also to be able to set up the source maps more easily. I already updated Netlify to use the new DSNs.


earn/src/index.tsx
```diff
@@ -11,8 +10,9 @@ import App from './App';
 if (process.env.REACT_APP_SENTRY_DSN) {
   Sentry.init({
     dsn: process.env.REACT_APP_SENTRY_DSN,
-    integrations: [new BrowserTracing()],
-
+    integrations: [new Sentry.BrowserTracing()],
+    // TODO: Automate this
```
haydenshively, line 14: I'm okay with merging this with the TODO, but I would like to hear what the plan is to address it in the future. Like in broad strokes how would that work?

haydenshively (approved):

earn/src/index.tsx
IanWoodard: Sentry has a GitHub Action we can use to automate this.

PR #409, opened 2023-04-30
Starting to remove our etherscan api dependency

IanWoodard:
The etherscan api has been a pain point for us as it has often been slow and even unresponsive at times. Moreover, it is not particularly scalable. This PR aims to start moving our code away from using etherscan and instead to using `eth_getLogs`. The one worry I have with this switch is issues arising from having too many logs and, thus possibly not getting the most updated data or even an error. That is why I chose to leave out the graph data fetching from this PR as it has the potential to grow to a size that may cause issues. Moreover, it does not block the rendering of the page meaning I am less worried about it taking a little longer. I do however hope to come back to this case in the future and properly address it.

I did my best to keep the logic roughly the same as before besides changing the etherscan API call with the `eth_getLogs`. I do plan on updating this logic more in the future, but I thought this would make for a good starting point.

### DIFF
--- a/earn/src/components/portfolio/LendingPairPeerCard.tsx
+++ b/earn/src/components/portfolio/LendingPairPeerCard.tsx
@@ -178,13 +178,13 @@ export default function LendingPairPeerCard(props: LendingPairPeerCardProps) {
       let lender1Logs: ethers.providers.Log[] = [];
       try {
         [lender0Logs, lender1Logs] = await Promise.all([
-          await provider.getLogs({
+          provider.getLogs({
             fromBlock: 0,
             toBlock: 'latest',
             address: selectedLendingPair.kitty0.address,
             topics: ['0xdcbc1c05240f31ff3ad067ef1ee35ce4997762752e3a095284754544f4c709d7'],
           }),
-          await provider.getLogs({
+          provider.getLogs({
             fromBlock: 0,
             toBlock: 'latest',
             address: selectedLendingPair.kitty1.address,
@@ -194,8 +194,9 @@ export default function LendingPairPeerCard(props: LendingPairPeerCardProps) {
       } catch (error) {
         console.error(error);
       }
-      if (lender0Logs == null || !Array.isArray(lender0Logs) || lender1Logs == null || !Array.isArray(lender1Logs))
+      if (lender0Logs.length === 0 && lender1Logs.length === 0) {
         return;
+      }
       let uniqueUsers = new Set<string>();
       const logs = [...lender0Logs, ...lender1Logs];
       logs.forEach((log: ethers.providers.Log) => {
--- a/earn/src/components/portfolio/LendingPairPeerCard.tsx
+++ b/earn/src/components/portfolio/LendingPairPeerCard.tsx
@@ -1,16 +1,16 @@
 import { useContext, useEffect, useMemo, useState } from 'react';
 
-import { AxiosResponse } from 'axios';
+import { ethers } from 'ethers';
 import { Dropdown, DropdownOption } from 'shared/lib/components/common/Dropdown';
 import { Display, Text } from 'shared/lib/components/common/Typography';
 import { Token } from 'shared/lib/data/Token';
 import { formatTokenAmount, roundPercentage } from 'shared/lib/util/Numbers';
 import styled from 'styled-components';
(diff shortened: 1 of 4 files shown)


earn/src/components/portfolio/LendingPairPeerCard.tsx
```diff
@@ -169,44 +170,37 @@ export default function LendingPairPeerCard(props: LendingPairPeerCardProps) {
...
-      let etherscanResultLender0: AxiosResponse<any, any> | null = null;
-      let etherscanResultLender1: AxiosResponse<any, any> | null = null;
+      let lender0Logs: ethers.providers.Log[] = [];
+      let lender1Logs: ethers.providers.Log[] = [];
       try {
-        [etherscanResultLender0, etherscanResultLender1] = await Promise.all([
-          etherscanRequestLender0,
-          etherscanRequestLender1,
+        [lender0Logs, lender1Logs] = await Promise.all([
+          await provider.getLogs({
```
haydenshively, line 181: Remove `await`?

earn/src/components/portfolio/LendingPairPeerCard.tsx
```diff
@@ -169,44 +170,37 @@ export default function LendingPairPeerCard(props: LendingPairPeerCardProps) {
...
+            fromBlock: 0,
+            toBlock: 'latest',
+            address: selectedLendingPair.kitty0.address,
+            topics: ['0xdcbc1c05240f31ff3ad067ef1ee35ce4997762752e3a095284754544f4c709d7'],
+          }),
+          await provider.getLogs({
```
haydenshively, line 187: Remove `await`

earn/src/components/portfolio/LendingPairPeerCard.tsx
```diff
@@ -169,44 +170,37 @@ export default function LendingPairPeerCard(props: LendingPairPeerCardProps) {
...
       } catch (error) {
         console.error(error);
       }
-      if (
-        etherscanResultLender0 == null ||
-        !Array.isArray(etherscanResultLender0.data.result) ||
-        etherscanResultLender1 == null ||
-        !Array.isArray(etherscanResultLender1.data.result)
-      )
+      if (lender0Logs == null || !Array.isArray(lender0Logs) || lender1Logs == null || !Array.isArray(lender1Logs))
```
haydenshively, line 197: When would these be null?

earn/src/data/LendingPair.ts
```diff
@@ -63,21 +61,20 @@ export async function getAvailableLendingPairs(
...
+      fromBlock: 7537163,
+      toBlock: 'latest',
+      address: ALOE_II_FACTORY_ADDRESS,
+      topics: ['0x3f53d2c2743b2b162c0aa5d678be4058d3ae2043700424be52c04105df3e2411'],
+    });
   } catch (e) {
     console.error(e);
   }
-  if (etherscanResult == null || !Array.isArray(etherscanResult.data.result)) return [];
+  if (logs == null || !Array.isArray(logs)) return [];
```
haydenshively, line 75: When would this be null?

earn/src/data/MarginAccount.ts
```diff
@@ -52,20 +51,23 @@ export type MarginAccount = {
...
+    logs = await provider.getLogs({
+      fromBlock: 0,
+      toBlock: 'latest',
+      address: ALOE_II_FACTORY_ADDRESS,
+      topics: [TOPIC0_CREATE_BORROWER_EVENT, null, `0x000000000000000000000000${userAddress.slice(2)}`],
+    });
+  } catch (e) {
+    console.error(e);
+  }
+  if (logs == null || !Array.isArray(logs)) return [];
```
haydenshively, line 68: ?

haydenshively (approved):

earn/src/components/portfolio/LendingPairPeerCard.tsx
IanWoodard: good catch

earn/src/components/portfolio/LendingPairPeerCard.tsx
IanWoodard: good catch

haydenshively (approved):

PR #416, opened 2023-05-03
An alternate way of fixing the markets page when not logged in

IanWoodard:
Similar to #411, but just disables the button so the user can't open the modal if they don't have a wallet connected.


haydenshively (approved):
Opacity doesn't seem to be working, but otherwise I like this solution much better.

PR #417, opened 2023-05-03
Additions to GoodNumbers before adding it to prime

IanWoodard:
Added some new functionality to GoodNumbers in preparation for adding it to prime. 

- added the base argument to the `fromDecimalString` and `fromDecimalBig` functions.
- added `one` function to allow the user to easily create a `GN` with a value of 1.
- added `isGteZero`, `isLtZero`, and `isLteZero` as they will be useful.
- added `areWithinNSigDigs` as we use this with numbers that will most likely become `GN`s.
- added tests to get full coverage.

### DIFF
--- a/shared/src/data/GoodNumber.test.ts
+++ b/shared/src/data/GoodNumber.test.ts
@@ -197,18 +197,18 @@ describe('GoodNumber', () => {
     it('should return true if two values are within N significant digits', () => {
       const a = GN.fromDecimalString('1', 18);
       const b = GN.fromDecimalString('1.000000000000000001', 18);
-      expect(GN.areWithinNSigDigs(a, b, 18)).toBe(true);
+      expect(GN.firstNSigDigsMatch(a, b, 18)).toBe(true);
       const c = GN.fromDecimalString('1.0235', 18);
       const d = GN.fromDecimalString('1.02', 18);
-      expect(GN.areWithinNSigDigs(c, d, 3)).toBe(true);
+      expect(GN.firstNSigDigsMatch(c, d, 3)).toBe(true);
     });
     it('should return false if two values are not within N significant digits', () => {
       const a = GN.fromDecimalString('1', 18);
       const b = GN.fromDecimalString('1.000000000000000001', 18);
-      expect(GN.areWithinNSigDigs(a, b, 19)).toBe(false);
+      expect(GN.firstNSigDigsMatch(a, b, 19)).toBe(false);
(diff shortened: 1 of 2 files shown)


shared/src/data/GoodNumber.test.ts
```diff
@@ -308,6 +373,28 @@ describe('GoodNumber', () => {
     });
   });
 
+  describe('one', () => {
+    it('should return a one value', () => {
+      expect(GN.one(18).toString(GNFormat.INT)).toEqual('1');
```
haydenshively, line 378: This should equal '1000000000000000000' not '1'

shared/src/data/GoodNumber.ts
```diff
@@ -149,6 +173,10 @@ export class GN {
     return a.lt(b) ? a : b;
   }
 
+  static areWithinNSigDigs(a: GN, b: GN, n: number): boolean {
```
haydenshively, line 176: I know you're just taking the name from elsewhere, but I kinda want to take this opportunity to call it "eqUpToNSigDigs" or "doFirstNSigDigsMatch" -- I feel these more accurately describe the condition we're checking.

shared/src/data/GoodNumber.ts
```diff
@@ -149,6 +173,10 @@ export class GN {
...
+    return a.int.prec(n).eq(b.int.prec(n));
```
haydenshively, line 177: Honestly didn't realize this `prec` function was a thing. To match Solidity more closely, one might argue we should be using it after every `mul` operation (since `Big.times` doesn't round automatically). Something to consider for future PRs.

haydenshively (commented):
A few things to fix. Hopefully doesn't mess with your broader changes.

shared/src/data/GoodNumber.ts
```diff
@@ -271,6 +299,10 @@ export class GN {
     return new GN('0', decimals, base);
   }
 
+  static one(decimals: number, base: 2 | 10 = 10) {
+    return new GN('1', decimals, base);
```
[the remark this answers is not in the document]
IanWoodard: I assume you meant `return new GN(scalerFor(base, decimals), decimals, base)`, but fixed.

haydenshively (dismissed):

shared/src/data/GoodNumber.ts
IanWoodard: Updated it to `firstNSigDigsMatch`

haydenshively (approved):

PR #421, opened 2023-05-05
Removing referral modal and adding bridge modal

IanWoodard:
Adding the foundation for the new bridge modal and removing the referral modal.
image
image


haydenshively (approved):
LGTM

PR #423, opened 2023-05-07
Adding the socket widget to the bridge modal

IanWoodard:
Integrating the socket widget with our UI. The socket widget is customized to fit our color scheme and fonts. Additionally, We limit the chains to those currently supported by the Aloe-ii protocol as well as Ethereum mainnet. The default source chain is set to mainnet and the default destination chain is set to whatever chain the user is currently connected to. When we launch on mainnet there may be a few additional changes that will need to be made. Additionally, we attempt to set the default destination token to whatever token the user has selected. This seems to work the first time the user opens the modal, but subsequent times in the same session result in the token being reset to USDC. I do not think this is within our control and is probably just a result of the widget's functionality, either intentionally or unintentionally.

image
image
image


earn/src/components/portfolio/modal/BridgeModal.tsx
```diff
@@ -1,18 +1,62 @@
...
+      <Bridge
+        provider={provider}
+        API_KEY={process.env.REACT_APP_SOCKET_API_KEY!}
+        customize={{
+          primary: 'rgb(13, 23, 30)',
+          secondary: 'rgb(26, 41, 52)',
+          onInteractive: 'rgb(255, 255, 255)',
+          text: 'rgb(255, 255, 255)',
+          secondaryText: 'rgb(255, 255, 255)',
+          interactive: 'rgb(26, 41, 52)',
```
haydenshively, line 50: Should these be constants?

haydenshively (approved):
LGTM 🚀

earn/src/components/portfolio/modal/BridgeModal.tsx
IanWoodard: We can address this in the future with the other colors that should be constants.

PR #424, opened 2023-05-07
Adding useProminentColor hook

IanWoodard:
Our current implementation for computing the prominent color was inefficient as it would repeat work a lot and wasn't very uniform across components. The `useProminentColor` hook aims to address these issues by making an efficient and easy-to-use hook for computing the prominent color of an icon. This hook is not aimed to address the instances where we do larger scale computation on multiple token icons at once, as those instances don't suffer the same inefficiencies, and this hook would not help in those cases.


earn/src/data/hooks/UseProminentColor.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useState } from 'react';
+
+import { getProminentColor } from '../../util/Colors';
+
+export default function useProminentColor(path: string) {
```
haydenshively, line 5: Could be more specific than `path`

haydenshively (approved):
LGTM

haydenshively (approved):

PR #432, opened 2023-05-17
Bug fixes for Prime Add Liquidity Card

IanWoodard:
Fixing various bugs with the Add Liquidity Card.

### DIFF
--- a/prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
+++ b/prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
@@ -62,6 +62,7 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
   // MARK: state for user inputs
   const [localIsAmount0UserDefined, setLocalIsAmount0UserDefined] = useState(false);
   const [localTokenAmounts, setLocalTokenAmounts] = useState<readonly [string, string]>(['', '']);
+  const [localLiquidity, setLocalLiquidity] = useState<JSBI>(JSBI.BigInt(0));
 
   // MARK: wagmi hooks
   const provider = useProvider({ chainId: activeChain.id });
@@ -227,6 +228,7 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
 
     setLocalTokenAmounts([amount0Str, amount1Str]);
     setLocalIsAmount0UserDefined(isToken0);
+    setLocalLiquidity(liquidity);
 
     callbackWithFullResults(isToken0Selected, amount0Str, amount1Str, lower, upper, liquidity);
   }
@@ -436,7 +438,14 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
             token1={token1}
             isToken0Selected={isToken0Selected}
             setIsToken0Selected={(value: boolean) => {
-              callbackWithFullResults(value, previousAmount0Str, previousAmount1Str, previousLower, previousUpper);
+              callbackWithFullResults(
+                value,
+                previousAmount0Str,
+                previousAmount1Str,
+                previousLower,
+                previousUpper,
+                localLiquidity
+              );
             }}
           />
         )}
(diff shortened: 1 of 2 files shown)


prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
```diff
@@ -305,9 +309,25 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
...
+        const priceAboveNearestTick = numericPriceRatioGN(
+          tickToPrice(nearestTick + tickInfo.tickSpacing),
+          token0.decimals,
+          token1.decimals,
+          !isToken0Selected
+        );
+        const closestTick =
+          Math.abs(price - priceAtNearestTick) < Math.abs(price - priceAboveNearestTick)
+            ? nearestTick
+            : nearestTick + tickInfo.tickSpacing;
```
haydenshively, line 327: I don't understand what this is for. We already define `nearestTick` -- why is `closestTick` necessary on top of that? If anything, the functions we use to compute `nearestTick` should be fixed, instead of adding this extra logic here.

prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
```diff
@@ -357,9 +377,25 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
...
```
haydenshively, line 395: Same.

prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
```diff
@@ -62,6 +62,7 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
   // MARK: state for user inputs
   const [localIsAmount0UserDefined, setLocalIsAmount0UserDefined] = useState(false);
   const [localTokenAmounts, setLocalTokenAmounts] = useState<readonly [string, string]>(['', '']);
+  const [localLiquidity, setLocalLiquidity] = useState<JSBI>(JSBI.BigInt(0));
```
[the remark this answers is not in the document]
IanWoodard: We talked about this offline about a week and a half back. It fixes the issue where the user would change the token in an erroring add liquidity action. The error would disappear since it would default to zero liquidity (until updated) since we did not store the liquidity.

prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
IanWoodard: Could you take a stab at fixing the underlying function? I spent a bit of time playing with it to no avail.

haydenshively: Definitely. Can you describe what's wrong with it so I know what to fix?

prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
```diff
@@ -302,7 +306,7 @@ export default function UniswapAddLiquidityActionCard(props: ActionCardProps) {
 
         if (!isToken0Selected) price = 1.0 / price;
         const nearestTick = roundDownToNearestN(
-          priceToTick(price, token0.decimals, token1.decimals),
+          priceToClosestTick(price, token0.decimals, token1.decimals),
           tickInfo.tickSpacing
         );
```
haydenshively, line 311: I'm glad you were able to factor out your logic into `priceToClosestTick`, but the reasoning for this still isn't clear to me. We're already `floor`ing or `ceil`ing the resultant tick value with `roundDownToNearestN` and `roundUpToNearestN`, so why is the `priceToClosestTick` thing necessary in the first place?

Like aside from whether this is the right fix, what bug are you addressing in the first place?

haydenshively (commented):
Still confused by the rounding stuff. Otherwise looks good.

prime/src/components/borrow/actions/UniswapAddLiquidityActionCard.tsx
IanWoodard: priceToTick was having issues with rounding, causing it to oftentimes return the tick above/below the actual closest tick. This caused issues with changing and updating ticks not working properly, as it sometimes skips over valid ticks (calculated by incrementing/decrementing the ticks) due to rounding. As for the outermost rounding, I am not sure this is doing much, but you can enlighten me on that as I believe you added it back in the day.

haydenshively: #436 covers it

haydenshively (dismissed):

PR #433, opened 2023-05-18
Adding 80% max button to borrow action card

IanWoodard:
Instead of having the max button populate the max amount a user can borrow, we instead populate it with 80% of the max to give the user a large portion of their potential borrows without putting them way too close to being liquidatable.


prime/src/components/borrow/actions/AloeBorrowActionCard.tsx
```diff
@@ -78,6 +78,8 @@ export function AloeBorrowActionCard(prop: ActionCardProps) {
   const max =
     selectedTokenOption.value === TokenType.ASSET0 ? GN.min(allowed0, available0) : GN.min(allowed1, available1);
   const maxString = max.toString(GNFormat.DECIMAL);
+  const maxEightyPercent = max.recklessMul(80).recklessDiv(100);
```
haydenshively, line 81: You can just recklessMul by 0.8 now that we changed GoodNumber.

haydenshively (approved):
I don't love that it labels the actual maximum as a "balance", since the amount you can borrow isn't really your balance. It's more of a limit. But the 80% logic and UI is good.

IanWoodard:
> I don't love that it labels the actual maximum as a "balance", since the amount you can borrow isn't really your balance. It's more of a limit. But the 80% logic and UI is good.

I agree, but that is out of the scope of this PR. I did note it in my list of things to do.

PR #434, opened 2023-05-18
Adding 80% max button to withdraw modal

IanWoodard:
Whenever the user has outstanding borrows, withdrawing max can put them very close to insolvency. To mitigate this, if the user has outstanding borrows, we populate the max amount with 80% of the max to give them a bit of leeway. If the user does not have outstanding borrows, the max is as it was previously.

### DIFF
--- a/prime/src/components/borrow/actions/AloeWithdrawActionCard.tsx
+++ b/prime/src/components/borrow/actions/AloeWithdrawActionCard.tsx
@@ -70,8 +70,13 @@ export function AloeWithdrawActionCard(prop: ActionCardProps) {
     token1.decimals
   );
 
+  const hasOutstandingBorrows = Object.values(accountState.liabilities).some((liability) => liability.isGtZero());
+
   const max = selectedTokenOption.value === TokenType.ASSET0 ? allowed0 : allowed1;
-  const maxString = max.toString(GNFormat.DECIMAL);
+  const eightyPercentMax = max.recklessMul(80).recklessDiv(100);
+  const maxString = hasOutstandingBorrows
+    ? eightyPercentMax.toString(GNFormat.DECIMAL)
+    : max.toString(GNFormat.DECIMAL);
 
   return (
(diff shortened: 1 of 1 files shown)


prime/src/components/borrow/actions/AloeWithdrawActionCard.tsx
```diff
@@ -70,8 +70,14 @@ export function AloeWithdrawActionCard(prop: ActionCardProps) {
     token1.decimals
   );
 
+  const hasOutstandingBorrows = Object.values(accountState.liabilities).some((liability) => liability.isGtZero());
+
   const max = selectedTokenOption.value === TokenType.ASSET0 ? allowed0 : allowed1;
-  const maxString = max.toString(GNFormat.DECIMAL);
+  const eightyPercentMax = max.recklessMul(80).recklessDiv(100);
```
haydenshively, line 76: Could just do 0.8

prime/src/components/borrow/actions/AloeWithdrawActionCard.tsx
```diff
@@ -70,8 +70,14 @@ export function AloeWithdrawActionCard(prop: ActionCardProps) {
...
+  const maxString = hasOutstandingBorrows
+    ? eightyPercentMax.toString(GNFormat.DECIMAL)
+    : max.toString(GNFormat.DECIMAL);
```
haydenshively, line 79:
```suggestion
  const maxString = (hasOutstandingBorrows ? eightyPercentMax : max).toString(GNFormat.DECIMAL);
```

haydenshively (dismissed):

prime/src/components/borrow/actions/AloeWithdrawActionCard.tsx
IanWoodard: While your suggestion is more condensed, I find having it on separate lines is more readable.

PR #436, opened 2023-05-19
Added invalid action message

IanWoodard:
Added a message at the bottom of the actions indicating one or more invalid action(s). There is also a button named "View Problem" that, when clicked, scrolls the manage account widget to the first invalid action.
image


prime/src/components/borrow/ManageAccountWidget.tsx
```diff
@@ -176,6 +193,11 @@ export default function ManageAccountWidget(props: ManageAccountWidgetProps) {
   const [errorMsg, setErrorMsg] = useState<string | undefined>(undefined);
   // modals
   const [showAddActionModal, setShowAddActionModal] = useState(false);
+  const actionCardRefs = useRef<(HTMLDivElement | null)[]>([]);
+
+  useEffect(() => {
+    actionCardRefs.current = actionCardRefs.current.slice(0, activeActions.length);
```
haydenshively, line 199: I don't understand how this works and what `current` refers to, but I'll trust you on it.

haydenshively: Is this just trimming down the array when actions are removed?

haydenshively (approved):

prime/src/components/borrow/ManageAccountWidget.tsx
IanWoodard: It synchronizes the length of actionCardRefs (used for scrolling to specific action cards) to the number of active actions. `current` is a special property of the `useRef` hook.

PR #439, opened 2023-05-22
Moving margin account to shared

IanWoodard:
Builds off of #438. Moving additional logic that is shared between apps to the shared directory.


haydenshively (approved):
LGTM

PR #443, opened 2023-05-28
Adding more compact token inputs to action cards

IanWoodard:
As the title suggests, I am adding a more compact input for the token amounts on the action cards. This component is similar to the one used in some of the newer portfolio modals on earn, but with slight modifications to work with `TokenType` rather than `Token`.
image


haydenshively (approved):
Yesss LGTM

PR #444, opened 2023-05-29
Cleaning up symbol usage since it is no longer optional

IanWoodard:
Currently, we have a lot of legacy optional usage of token symbols since they used to be optional. Since it is no longer an optional property, I updated the usage to reflect that.


haydenshively (approved):
Good call

PR #446, opened 2023-05-30
Simplify priceToClosestTick by using JS numbers

haydenshively:
@IanWoodard's math for this is correct, but quite verbose. His logic did the following:

- convert price to sqrtPriceX96
- use TickMath to convert sqrtPriceX96 to an integer tick _(this floors it)_
- convert that tick back to a price
- compare with original and +/- 1 as necessary to get the _actual_ nearest tick (undo-ing the floor operation from before)

Here, I instead compute a floating point tick and then just `Math.round` it. Output of the function is unchanged.

### DIFF
--- a/prime/src/util/Uniswap.ts
+++ b/prime/src/util/Uniswap.ts
@@ -260,39 +260,8 @@ export function tickToPrice(tick: number): GN {
  */
 export function priceToClosestTick(price0In1: number, token0Decimals: number, token1Decimals: number): number {
   const decimalDiff = token0Decimals - token1Decimals;
-  const priceX96 = new Big(price0In1).mul(BIGQ96).div(10 ** decimalDiff);
-
-  const sqrtPriceX48 = priceX96.sqrt();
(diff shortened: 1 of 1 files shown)


prime/src/util/Uniswap.ts
```diff
@@ -260,39 +260,8 @@ export function tickToPrice(tick: number): GN {
...
-    const priceAround = tickToNumericPrice(tickAround);
-
-    // If the price is closer to the tick above/below, use that tick instead
-    if (Math.abs(priceAround - price0In1) < Math.abs(priceAtTick - price0In1)) {
-      tickAtSqrtRatio = tickAround;
-    }
-  }
-
-  return tickAtSqrtRatio;
+  const tick = (Math.log10(price0In1) - decimalDiff) / Math.log10(1.0001);
```
haydenshively, line 263: $tick = log_{1.0001}(\frac{price0In1}{10^{decimalDiff}}) = \frac{log_{10}(\frac{price0In1}{10^{decimalDiff}})}{log_{10}(1.0001)} = \frac{log_{10}(price0In1) - decimalDiff}{log_{10}(1.0001)}$

IanWoodard (approved):
LGTM

PR #447, opened 2023-05-31
Fixing pie chart on prime to have correct categories

IanWoodard:
Currently, the pie chart on prime has uniswap and raw flipped for token1. This PR aims to fix that.


haydenshively (approved):
Good catch

PR #451, opened 2023-06-02
Reworking how we display health

IanWoodard:
Making the health displays consistent across prime. Adding more decimal points in the manage account widget health. Adding text coloring to the manage account widget health. One thing to note is that I did update the max health that the manage account widget displays (as "X+") from 3 to 5 (to match the accounts page); however, I did notice that the health bar relies on the health values to be between 0.5 and 3, so technically any value greater than three shows as the max on the health bar, but doesn't show up as "5+" until a value of 5 is reached. If this doesn't make sense, don't worry about it too much. I just feel it is more important to match between pages. I recommend trying it out for yourself and seeing how it feels. We could also just use "3+" everywhere as an alternate option. 

image
image
image

### DIFF
--- a/prime/src/util/Health.ts
+++ b/prime/src/util/Health.ts
@@ -1,9 +1,15 @@
+export const HEALTH_LOW_COLOR = 'rgba(235, 87, 87, 1)';
+export const HEALTH_MEDIUM_COLOR = 'rgba(242, 201, 76, 1)';
+export const HEALTH_HIGH_COLOR = 'rgba(0, 193, 67, 1)';
+export const HEALTH_LOW_THRESHOLD = 1.05;
+export const HEALTH_MEDIUM_THRESHOLD = 1.5;
+
 export function getHealthColor(health: number): string {
-  if (health <= 1.1) {
-    return '#EB5757';
-  } else if (health <= 1.25) {
-    return '#F2C94C';
+  if (health < HEALTH_LOW_THRESHOLD) {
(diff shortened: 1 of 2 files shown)


prime/src/util/Health.ts
```diff
@@ -1,9 +1,15 @@
+export const HEALTH_LOW_COLOR = 'rgba(235, 87, 87, 1)';
+export const HEALTH_MEDIUM_COLOR = 'rgba(242, 201, 76, 1)';
+export const HEALTH_HIGH_COLOR = 'rgba(0, 193, 67, 1)';
+export const HEALTH_LOW_THRESHOLD = 1.05;
+export const HEALTH_MEDIUM_THRESHOLD = 1.5;
```
haydenshively, line 5: This was 1.25 before. 1.5 is pretty high I think.

haydenshively (approved):

prime/src/util/Health.ts
IanWoodard: I updated it to match the health bar, which turns yellow at around 1.5

haydenshively: It's more of a greenish yellow at that point though. This is straight up yellow

IanWoodard: What if we met in the middle at around 1.35?

haydenshively: sgtm

PR #453, opened 2023-06-06
Simplifying uniswap position card logic

IanWoodard:
Refactoring the uniswap position list making it more straightforward. Additionally, I made it so that uniswap positions appear empty before loading rather than not showing at all to improve the loading appearance and mitigate layout shifting.


haydenshively (approved):
LGTM

PR #464, opened 2023-06-18
Adding sorting to the markets page (by apy)

IanWoodard:
Currently, we do not have any special sorting for the markets page. This PR changes that by sorting the lending cards by apy. The sorting is done by taking the combined average of each lender's apy in the pair.


haydenshively (approved):
Yay!

PR #466, opened 2023-06-22
Fixing inputs on prime

IanWoodard:
Fixing number inputs causing parsing errors with GN.


prime/src/components/borrow/actions/AloeAddMarginActionCard.tsx
```diff
@@ -80,7 +81,12 @@ export function AloeAddMarginActionCard(prop: ActionCardProps) {
         options={dropdownOptions}
         selectedOption={selectedTokenOption}
         maxAmount={maxString}
-        onChange={(value) => callbackWithFullResult(selectedToken, value)}
+        onChange={(value) => {
+          const output = formatNumberInput(value);
+          if (output != null) {
+            callbackWithFullResult(selectedToken, value);
+          }
+        }}
```
haydenshively, line 89: Seems like all of these should be done inside `callbackWithFullResult`. Is there a reason you did it down here?

haydenshively (approved):

PR #483, opened 2023-06-25
Fixing lend page filter bug

IanWoodard:
Fixing a bug where changing the filter settings while on a page besides the first can cause the app to crash.


earn/src/pages/LendPage.tsx
```diff
@@ -35,6 +35,7 @@ import {
 } from '../data/LendingPair';
 import { PriceRelayLatestResponse } from '../data/PriceRelayResponse';
 
+const MIN_PAGE_NUMBER = 1;
```
haydenshively, line 38: Interesting constant 🤔

haydenshively (approved):

PR #497, opened 2023-07-01
Only show modals when connected

IanWoodard:
Instead of potentially stacking up a bunch of open modals while disconnected, limit the opening of modals to connected accounts.


haydenshively (commented):
Does it really hurt to pre-load these though? We expect users to connect a wallet eventually, so they'll have to be loaded anyway

IanWoodard:
> Does it really hurt to pre-load these though? We expect users to connect a wallet eventually, so they'll have to be loaded anyway

The main goal of this fix is to disable the open modal buttons when the user's wallet is not connected to prevent multiple modals from poping up simultaneously when the user does connect their wallet. I also feel that it isn't great UX to have a modal you clicked on before connecting, suddenly open when you connect.

haydenshively (approved):

PR #502, opened 2023-07-04
Upgrading to wallet connect v2

IanWoodard:
Since wallet connect v1 has been sunset; we must upgrade to wallet connect v2. To do this, we need to upgrade our wagmi version to 0.12.16. This upgrade requires some changes to our usage of wagmi. I did my best to keep the changes minimal and focused on keeping things similar to before while still working with the new wagmi changes.

### DIFF
--- a/shared/src/components/WagmiProvider.tsx
+++ b/shared/src/components/WagmiProvider.tsx
@@ -1,5 +1,6 @@
 import { ethers } from 'ethers';
-import { WagmiConfig, chain, createClient, configureChains } from 'wagmi';
+import { WagmiConfig, createClient, configureChains } from 'wagmi';
+import { arbitrum, optimism, mainnet } from 'wagmi/chains';
 import { CoinbaseWalletConnector } from 'wagmi/connectors/coinbaseWallet';
 import { InjectedConnector } from 'wagmi/connectors/injected';
 import { WalletConnectConnector } from 'wagmi/connectors/walletConnect';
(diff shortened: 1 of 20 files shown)


shared/src/components/WagmiProvider.tsx
```diff
@@ -8,11 +9,12 @@ import { infuraProvider } from 'wagmi/providers/infura';
...
 
 import { DEFAULT_CHAIN } from '../data/constants/Values';
+import { ALL_CHAINS } from '../data/constants/ChainSpecific';
 
 function fallbackProvider({ chainId }: { chainId?: number }) {
-  const targetChain = Object.values(chain).find((v) => v.id === chainId) || DEFAULT_CHAIN;
+  const targetChain = ALL_CHAINS.find((v) => v.id === chainId) || DEFAULT_CHAIN;
   const config = {
-    ensAddress: targetChain.ens?.address,
+    // ensAddress: targetChain.ens?.address,
```
IanWoodard, line 17: This no longer works.

haydenshively: :(

shared/src/components/WagmiProvider.tsx
```diff
@@ -88,7 +89,7 @@ const client = createClient({
     }),
     new WalletConnectConnector({
       chains,
-      options: { qrcode: true },
+      options: { projectId: '3475bfb3ffc4cd0576fabe5d47efeafb' },
```
haydenshively, line 92: Should this be a secret?

haydenshively (approved):

shared/src/components/WagmiProvider.tsx
IanWoodard: Not particularly, I don't think. But I moved it to env vars in the next PR.

PR #523, opened 2023-07-21
Fixing uniswap table showing empty positions

IanWoodard:
Currently, we show "ghost" positions that aren't removable.


haydenshively:
Note that they are removable, as the user has complete control over what's in that Uniswap positions array. But since they contain no liquidity, we can safely ignore them and just overwrite them if they create new positions -- no need to explicitly remove.

PR #536, opened 2023-07-28
Adding initial boost page

IanWoodard:


earn/src/data/Uniswap.ts
```diff
@@ -8,13 +11,17 @@ import JSBI from 'jsbi';
...
 import { toBig } from 'shared/lib/util/Numbers';
+import { Address } from 'wagmi';
 
 import UniswapNFTManagerABI from '../assets/abis/UniswapNFTManager.json';
 import UniswapV3PoolABI from '../assets/abis/UniswapV3Pool.json';
 import { convertBigNumbersForReturnContexts } from '../util/Multicall';
 import { UNISWAP_NONFUNGIBLE_POSITION_MANAGER_ADDRESS } from './constants/Addresses';
 import { BIGQ96, Q96 } from './constants/Values';
 
+const FACTORY_ADDRESS = '0x1F98431c8aD98523631AE4a59f267346ea31F984';
```
haydenshively, line 22: Maybe call this `UNISWAP_FACTORY_ADDRESS` to differentiate it from our factory.

earn/src/pages/BoostPage.tsx
```diff
@@ -0,0 +1,209 @@
...
+        lower: tickLower,
+        upper: tickUpper,
+        liquidity: position.liquidity,
+      };
+
+      const [amount0, amount1] = getAmountsForLiquidity(uniswapPosition, currentTick, token0.decimals, token1.decimals);
+
+      const token0PerToken1 = sqrtRatioToPrice(toBig(sqrtPriceX96), token0.decimals, token1.decimals);
+      const amount0InTermsOfToken1 = amount0 * token0PerToken1;
+      const totalValue = amount0InTermsOfToken1 + amount1;
```
haydenshively, line 99: We have `GoodNumber` ways of doing this, no?

haydenshively (approved):

earn/src/data/Uniswap.ts
IanWoodard: This is inside of the uniswap class so I think it is pretty clear. We can update it in the future.

earn/src/pages/BoostPage.tsx
IanWoodard: I believe this code is taken from other places, but we can look into updating this in the future.

PR #563, opened 2023-08-07
Tie boost page into backend

haydenshively:
Description for f4d4edc23a286afc0e2fc73d24ce717c236da395:

> Our goal with Boost has been to make leveraged Uniswap positions a 1-click process. Before, users would have to (a) create a `Borrower` and (b) call `modify` with appropriate args to import a Uniswap position, borrow money, and then add the borrowed money to the position.
> 
> With the new BoostNFT contract, those steps are combined. Calling `mint` will generate a new `Borrower` and call `modify` all at once. After minting, the `Borrower` is permanently associated with the `id` of the newly-minted NFT.
> 
> In this PR, we do the following:
> - Figure out how many Boost NFTs the user has
> - Compute the `id` of each NFT
> - For every `id`, get the associated `Borrower`
> - Fetch all sorts of info about the `Borrower`(s). _This step is very similar to what we already do when fetching margin accounts._
> 
> With a bit of refactoring, we probably could have reused the existing margin accounts fetcher, but I decided to write a new one dedicated to Boost Borrowers. Downside is it's only setup to do multicall _within_ a given borrower, whereas the other one merges calls across all borrowers. We can improve this in the future.

Also includes #564 #565 #566 #567 and #572

Reviewing commit-by-commit will make more sense than all at once.

Remaining TODOs:
- [ ] Page description / help box at the top
- [ ] Button labels for the wrench + lightning bolt
- [ ] Withdrawing from a Boost NFT
- [ ] Calculating liquidation thresholds (for Boost NFTs show current, for Uniswap NFTs show hypothetical when importing)
- [ ] Displaying liquidation threshold icons on chart
- [ ] Displaying position bounds icons on chart
- [ ] Displaying earned fees instead of min/max prices
- [ ] Styling + buttons in the import modal


earn/src/data/Uniboost.ts
```diff
@@ -0,0 +1,206 @@
...
+import VolatilityOracleAbi from '../assets/abis/VolatilityOracle.json';
+import { ALOE_II_BORROWER_LENS_ADDRESS, ALOE_II_ORACLE_ADDRESS } from './constants/Addresses';
+import { Assets, Liabilities, MarginAccount } from './MarginAccount';
+import { UniswapPosition } from './Uniswap';
+
+const BOOST_NFT_ADDRESSES: { [chainId: number]: Address } = {
+  [optimism.id]: '0xA58eEcBd367334E742554ce6A2CC8b6863487ebB',
+  [arbitrum.id]: '0xA58eEcBd367334E742554ce6A2CC8b6863487ebB',
+  8453: '0xecED17C61971A32E28bc55537e8103ce3002d0dA',
+};
```
haydenshively, line 24: I put this here since I know you're working on dynamic addresses elsewhere.

IanWoodard (approved):

PR #564, opened 2023-08-07
Prep BoostCardInfo type to accept Uniswap NFTs -and- our NFTs

haydenshively:
This data type had a lot of redundant properties (i.e. ones that could be computed from other properties) so I turned it into a class. This will make things simpler later on, when we're trying to instantiate `BoostCardInfo`s not only for Uniswap NFTs, but also for Aloe NFTs.


earn/src/pages/BoostPage.tsx
```diff
@@ -123,60 +155,23 @@ export default function BoostPage() {
...
-
-      const amount0Percent = totalValue > 0 ? (amount0InTermsOfToken1 / totalValue) * 100 : 0;
-      const amount1Percent = totalValue > 0 ? (amount1 / totalValue) * 100 : 0;
-
-      const isInRange = uniswapPosition && currentTick >= uniswapPosition.lower && currentTick <= uniswapPosition.upper;
-
-      const isDeposit = Math.random() > 0.5; // TODO: figure out how to determine if this is a deposit or withdrawal
-
-      const poolAddress = computePoolAddress(position);
-      const currentPrice = sqrtRatioToPrice(toBig(sqrtPriceX96), token0.decimals, token1.decimals);
```
haydenshively, line 159: All of this is done in the class' helper methods now.

earn/src/pages/BoostPage.tsx
```diff
@@ -5,43 +5,75 @@ import { UniswapV3PoolABI } from 'shared/lib/abis/UniswapV3Pool';
...
-  color1: string;
-  minTick: number;
-  maxTick: number;
-  currentTick: number;
-};
+export enum BoostCardType {
+  UNISWAP_NFT,
+  BOOST_NFT,
+  BOOST_NFT_GENERALIZED,
+}
```
haydenshively, line 29: Switched to enum instead of the `isDeposit` boolean since there are currently 3 states and likely more in the future (ex: BOOST_NFT_LIQUIDATED)

PR #567, opened 2023-08-07
Show boosted positions on the Boost page, not just Uniswap NFTs

haydenshively:
We were already fetching UniswapNFTPositions and then sculpting them to fit the BoostCardInfo data type.

Here I keep that, but also add code to do the same for BoostNFTs.

Once all the data is fetched, I merge the `BoostCardInfo[]` of each type (Boosted and Uniswap) into one big array. Since they get fetched at different rates, I had to make the `uniqueId` a bit fancier to make sure React knew which array indices to re-render.


earn/src/data/Uniboost.ts
```diff
@@ -39,9 +39,25 @@ export class BoostCardInfo {
...
+    const uniswapValue = getValueOfLiquidity(this.position, this.currentTick, this.token1.decimals);
+
+    // Compute total debt
+    const debt0 = this.borrower.liabilities.amount0 - this.borrower.assets.token0Raw;
+    const debt1 = this.borrower.liabilities.amount1 - this.borrower.assets.token1Raw;
+    const price = tickToPrice(this.currentTick, this.token0.decimals, this.token1.decimals, true);
+    const debtValue = debt0 * price + debt1;
+
+    return uniswapValue / (uniswapValue - debtValue);
+  }
```
haydenshively, line 59: Got ahead of myself here, this isn't actually used yet.

earn/src/pages/BoostPage.tsx
```diff
@@ -57,88 +84,131 @@ export default function BoostPage() {
...
+  }, [uniswapNFTPositions, slot0Data]);
+
+  /*//////////////////////////////////////////////////////////////
+                           COMPUTE COLORS
+  //////////////////////////////////////////////////////////////*/
+  useEffect(() => {
+    (async () => {
+      const merged = boostedCardInfos.concat(uniswapCardInfos);
+      const tokenAddresses = Array.from(
+        new Set(merged.flatMap((cardInfo) => [cardInfo.token0.logoURI, cardInfo.token1.logoURI])).values()
```
haydenshively, line 138: Since many tokens will be repeated, I used a Set here to make sure we only compute prominent color for each token once.

PR #574, opened 2023-08-12
Create chain dependent state hook for fast chain switching

haydenshively:
Basically just keeps track of which chain a given piece of state belongs to so that we don't have to refresh the page. In theory we could use cached data and not have to re-fetch anything after a chain has been loaded once, but right now it'll still refetch (_not_ refresh) every time you switch chains.

Expect a few contractRead errors on the Portfolio Page due to the modals referencing stale data (tokens that exist on one chain but not another). They're ugly in the console but it figures itself out.

Also, I counted the number of renders before and after this change and it's the same (or +1 in some cases). So I think it's safe to use.

We could merge this to master if I take out the BoostPage changes.


shared/src/components/navbar/ChainSelector.tsx
```diff
@@ -49,20 +48,21 @@ export default function ChainSelector(props: ChainSelectorProps) {
...
-    onSuccess: () => {
+    onMutate(args) {
+      console.debug('About to switch chains', args);
+    },
+    onSuccess: (chain) => {
+      setActiveChain(chain);
       setShouldAttemptToSwitchNetwork(false);
       setIsOpen(false);
       setPendingChainOption(undefined);
-      navigate(0);
```
haydenshively, line 65: Taking this out is what gives us the speed

shared/src/data/hooks/UseChainDependentState.ts
```diff
@@ -0,0 +1,29 @@
+import { useCallback, useMemo, useRef, useState } from 'react';
+
+export function useChainDependentState<S>(initialState: S, chainId: number) {
```
haydenshively, line 3: Drop-in replacement for React's `useState`

earn/src/components/portfolio/modal/WithdrawModal.tsx
```diff
@@ -83,9 +83,8 @@ export default function WithdrawModal(props: WithdrawModalProps) {
   const account = useAccount();
 
   const filteredPairs = lendingPairs.filter((p) => doesLendingPairContainToken(p, selectedToken));
-  if (filteredPairs.length === 0) throw new Error(`${selectedToken.symbol} isn't part of any lending pair`);
-  const selectedPair = filteredPairs.at(selectedPairIdx) ?? filteredPairs[0];
-  const lender = selectedPair.token0.equals(selectedToken) ? selectedPair.kitty0 : selectedPair.kitty1;
+  const selectedPair = filteredPairs.at(selectedPairIdx);
+  const lender = selectedPair?.token0.equals(selectedToken) ? selectedPair.kitty0 : selectedPair?.kitty1;
   const amount = GN.fromDecimalString(inputValue[0] || '0', selectedToken.decimals);
```
haydenshively, line 88: Ideally wouldn't have had to change this file, but there was some bad code in here from last time I touched it

earn/src/pages/BoostPage.tsx
```diff
@@ -26,10 +27,60 @@ export default function BoostPage() {
...
+            DEFAULT_COLOR1,
+            res.uniswapPosition,
+            res.uniswapFees,
+            res.borrower
+          );
+        })
+      );
+
+      if (mounted) setBoostedCardInfos(fetchedInfos);
+    })();
```
haydenshively, line 71: Tbh idk why this is showing up on this diff. Should've already been on master.

IanWoodard (approved):

PR #576, opened 2023-08-12
Stop mixing activeChain with useProvider on BoostPage

haydenshively:
Don't worry about this one yet, diff will get cleaned up once we merge some other stuff.


earn/src/pages/BoostPage.tsx
```diff
@@ -35,13 +35,18 @@ export default function BoostPage() {
                       FETCH BOOSTED CARD INFOS
   //////////////////////////////////////////////////////////////*/
   useEffect(() => {
+    let mounted = true;
```
haydenshively, line 38: @IanWoodard Certain online tutorials told me that the `mounted` flag you always use isn't necessary. They were wrong. Since it is in fact necessary and we use it everywhere, I'm thinking a `useAsyncEffect` hook would be really useful.

Some kind of generic wrapper around `useEffect` where we can pass an async lambda and the state update callback, and it automatically creates + handles the `mounted` conditional. Does that seem useful and/or doable?

IanWoodard: Haha. It isn't always needed, but it makes React happy. I believe they are doing away with the need for it in future versions of React. It could be useful to have a useAsyncEffect hook; I can look into adding one. I'll put it on my to-do list.

IanWoodard (approved):

PR #578, opened 2023-08-14
Updating the import modal

IanWoodard:
Adding more functionality and improving upon the import modal.
**Added**:
- a hypothetical boost in card
- summary section
- a more polished (but still WIP slider)
- a more polished button setup

image


earn/src/components/boost/ImportModal.tsx
```diff
@@ -175,77 +259,147 @@ export default function ImportModal(props: ImportModalProps) {
...
-      // TODO: prob want to reload the page in here. keep the console.debug if possible (for consistency since we have this style of thing elsewhere)
-      console.debug('Mint transaction successful!', data);
+    onSuccess() {
+      navigate(0);
     },
   });
 
-  // TODO: you can remove these console.logs obviously
-  let state: ImportModalState;
+  let state: ImportModalState | undefined;
```
haydenshively, line 267: Why/when would this be undefined?

earn/src/components/boost/ImportModal.tsx
```diff
@@ -73,6 +109,46 @@ enum ImportModalState {
   ASKING_USER_TO_MINT,
 }
 
+function getButtonState(state?: ImportModalState) {
+  switch (state) {
+    case ImportModalState.FETCHING_DATA:
+      return {
+        disabled: true,
+        label: 'Loading...',
+      };
```
haydenshively: imo these would be more readable on one line, and I'm pretty sure our Prettier config is fine with that

earn/src/components/boost/ImportModal.tsx
```diff
@@ -83,16 +159,23 @@ export type ImportModalProps = {
...
 
   const nftTokenId = ethers.BigNumber.from(cardInfo?.nftTokenId || 0);
   const initializationData = useMemo(() => {
     if (!cardInfo) return undefined;
     const { position } = cardInfo;
     return ethers.utils.defaultAbiCoder.encode(
-      ['uint256', 'int24', 'int24', 'uint128', 'uint8'],
-      [cardInfo.nftTokenId, position.lower, position.upper, position.liquidity.toString(10), boostFactor]
+      ['uint256', 'int24', 'int24', 'uint128', 'uint24'],
+      [cardInfo.nftTokenId, position.lower, position.upper, position.liquidity.toString(10), boostFactor * 10000]
```
haydenshively, line 178: You might want to do `(boostFactor * 10000).toFixed(0)` as future proofing.

earn/src/components/boost/ImportModal.tsx
```diff
@@ -110,15 +193,16 @@ export default function ImportModal(props: ImportModalProps) {
     chainId: activeChain.id,
     enabled: enableHooks,
   });
-  const shouldWriteManager = !isFetchingManager && !!manager && manager !== ALOE_II_BOOST_MANAGER_ADDRESS;
-  const shouldMint = !isFetchingManager && !!initializationData && manager === ALOE_II_BOOST_MANAGER_ADDRESS;
+  const shouldWriteManager =
+    !isFetchingManager && !!manager && manager !== ALOE_II_BOOST_MANAGER_ADDRESS[activeChain.id];
+  const shouldMint =
+    !isFetchingManager && !!initializationData && manager === ALOE_II_BOOST_MANAGER_ADDRESS[activeChain.id];
```
haydenshively, line 199: To simplify these, maybe first define another boolean like `const managerIsCorrect = manager === ALOE_II_BOOST_MANAGER_ADDRESS[activeChain.id]`.

earn/src/components/boost/ImportModal.tsx
```diff
@@ -175,77 +259,147 @@ export default function ImportModal(props: ImportModalProps) {
...
+
   // Generate labels for input range (slider)
   const labels: string[] = [];
   for (let i = BOOST_MIN; i <= BOOST_MAX; i += 1) {
-    labels.push(`${i.toFixed(0)}x`);
+    if (i % 2 !== 0) {
+      labels.push(`${i.toFixed(0)}x`);
+    } else {
+      labels.push('');
+    }
```
haydenshively, line 291: I don't think you need to define a label for every tick. You can just do it for the ticks we actually want, and then create fewer `<options>` down below.

earn/src/components/boost/ImportModal.tsx
```diff
@@ -175,77 +259,147 @@ export default function ImportModal(props: ImportModalProps) {
...
   }
 
+  const updatedCardInfo: BoostCardInfo | undefined = useMemo(() => {
+    if (!cardInfo) return undefined;
+    const { position } = cardInfo;
+    const updatedLiquidity = JSBI.multiply(position.liquidity, JSBI.BigInt(boostFactor));
```
haydenshively, line 297: If `boostFactor` has decimals I don't think this works.

earn/src/components/boost/ImportModal.tsx
```diff
@@ -175,77 +259,147 @@ export default function ImportModal(props: ImportModalProps) {
...
+      cardInfo.cardType,
+      cardInfo.nftTokenId,
+      cardInfo.uniswapPool,
+      cardInfo.currentTick,
+      cardInfo.token0,
+      cardInfo.token1,
+      cardInfo.color0,
+      cardInfo.color1,
+      {
+        ...cardInfo.position,
```
haydenshively, line 308: You defined `const {position}` above, do you want to use it?

earn/src/components/boost/ImportModal.tsx
```diff
@@ -175,77 +259,147 @@ export default function ImportModal(props: ImportModalProps) {
...
+        liabilities: {
+          amount0: cardInfo.amount0() * (boostFactor - 1),
+          amount1: cardInfo.amount1() * (boostFactor - 1),
+        },
+        feeTier: FeeTier.INVALID,
+        sqrtPriceX96: new Big(0),
+        health: 0,
+        lender0: '0x',
+        lender1: '0x',
+        iv: 0,
```
haydenshively, line 332: We may need sqrtPriceX96 and iv for liquidation thresholds, but out of scope for now.

earn/src/components/boost/ImportModal.tsx
```diff
@@ -175,77 +259,147 @@ export default function ImportModal(props: ImportModalProps) {
...
+    } else {
+      labels.push('');
+    }
   }
 
+  const updatedCardInfo: BoostCardInfo | undefined = useMemo(() => {
+    if (!cardInfo) return undefined;
+    const { position } = cardInfo;
+    const updatedLiquidity = JSBI.multiply(position.liquidity, JSBI.BigInt(boostFactor));
+    return new BoostCardInfo(
```
haydenshively, line 298: You could write a static factory method on `BoostCardInfo` that allows for usage like:

```
return BoostCardInfo.from(cardInfo, {
  marginAccount: {...},
  position: {...},
})
```

earn/src/components/boost/ImportModal.tsx
```diff
@@ -175,77 +259,147 @@ export default function ImportModal(props: ImportModalProps) {
...
+                  if (state === ImportModalState.READY_TO_APPROVE) {
+                    writeManager?.();
+                  } else if (state === ImportModalState.READY_TO_MINT) {
+                    mint?.();
+                  }
+                }}
+              >
+                {buttonState.label}
+              </FilledStylizedButton>
+              <Text size='XS' color={TERTIARY_COLOR} className='w-full mt-2'>
```
haydenshively, line 394: On the portfolio page, it looks like this text is usually left aligned, but here it's centered.

haydenshively (dismissed):
LGTM, just nits

earn/src/components/boost/ImportModal.tsx
IanWoodard: It has no default value and all of the places that set it are within if statements (no else)

haydenshively: Default should probably be the loading state then

IanWoodard: There is no "loading state", we just treat no state as loading state which is technically the same thing.

earn/src/components/boost/ImportModal.tsx
IanWoodard: I do this to add evenly spaced ticks that also show only the odd-numbered ticks to avoid displaying too much. I prefer this since it also shows tick marks where it doesn't show a label which not having the option does not.

haydenshively: Ah I see, didn't realize it was tied to the existence of tick marks.

earn/src/components/boost/ImportModal.tsx
IanWoodard: punting

earn/src/components/boost/ImportModal.tsx
IanWoodard: Punting

PR #593, opened 2023-08-21
Adding APR to boost import modal

IanWoodard:
image


earn/src/pages/BoostPage.tsx
```diff
@@ -227,14 +254,16 @@ export default function BoostPage() {
...
       </div>
-      <ImportModal
-        isOpen={selectedPosition !== null && selectedCardInfo?.cardType === BoostCardType.UNISWAP_NFT}
-        uniqueId={selectedCardInfo ? getUniqueId(selectedCardInfo) : ''}
-        setIsOpen={() => {
-          setSelectedPosition(null);
-        }}
-        cardInfo={selectedCardInfo}
-      />
+      {selectedCardInfo !== undefined && (
```
haydenshively, line 257: Why is this necessary for the ImportModal but not the ManageBoostModal?

earn/src/data/Uniboost.ts
```diff
@@ -36,6 +36,8 @@ export class BoostCardInfo {
     public readonly currentTick: number,
     public readonly token0: Token,
     public readonly token1: Token,
+    public readonly lender0: Address,
+    public readonly lender1: Address,
```
haydenshively, line 40: By adding these to the class, are you generating extra re-renders? i.e., all graphs are rendering as before and liquidation thresholds are calculated as before... but then they're rendered/calculated _again_ once `lender0` and `lender1` come in?

haydenshively: If this is the case, we should track lender0 and lender1 in some other data type.

earn/src/components/boost/ImportModal.tsx
```diff
@@ -138,15 +150,43 @@ export default function ImportModal(props: ImportModalProps) {
...
+      );
+      if (isMounted) {
+        setMarketInfo(marketInfo);
+      }
+    }
+    fetchMarketInfo();
+    return () => {
+      isMounted = false;
+    };
+  }, [activeChain.id, cardInfo, provider, setMarketInfo]);
```
haydenshively, line 188: This could probably be done with wagmi hooks, but not a big deal either way.

haydenshively (approved):

earn/src/pages/BoostPage.tsx
IanWoodard: I haven't updated the `ManageBoostModal` yet. It just allows us to not have undefined `cardInfo` in the modal.

earn/src/components/boost/ImportModal.tsx
IanWoodard: I am just re-using old code here.

earn/src/data/Uniboost.ts
IanWoodard: Good catch

IanWoodard: Actually, the way we currently do things, there should be no additional chart renders as we have a guard clause that ensures marketData is fetched before constructing the BoostCardInfos.

PR #634, opened 2023-09-16
Adding withdraw ante modal

IanWoodard:
There is still some more work to be done potentially with regard to when the button is disabled, but this is meant as a starting point.
image

### DIFF
--- a/earn/src/pages/BorrowPage.tsx
+++ b/earn/src/pages/BorrowPage.tsx
@@ -16,6 +16,7 @@ import {
   ALOE_II_ORACLE_ADDRESS,
 } from 'shared/lib/data/constants/ChainSpecific';
 import { GetNumericFeeTier } from 'shared/lib/data/FeeTier';
+import { GN } from 'shared/lib/data/GoodNumber';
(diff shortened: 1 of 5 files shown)


earn/src/pages/BorrowPage.tsx
```diff
@@ -481,6 +493,12 @@ export default function BorrowPage() {
   const baseEtherscanUrl = getEtherscanUrlForChain(activeChain);
   const selectedMarginAccountEtherscanUrl = `${baseEtherscanUrl}/address/${selectedMarginAccount?.address}`;
 
+  const isUnableToWithdrawAnte =
+    Object.values(selectedMarginAccount?.liabilities ?? {}).some((liability) => {
+      return liability > 0;
+    }) ||
+    (accountEtherBalance?.isZero() ?? true);
```
haydenshively, line 500: A tad hard to read

haydenshively (approved):

earn/src/pages/BorrowPage.tsx
haydenshively: I would split into 2 booleans, `accountHasEther` and `canWithdrawAnte`

PR #643, opened 2023-09-24
Update BalanceSheet for new probe prices method

haydenshively:
Updated the probe prices computation. Empirically the rest of the BalanceSheet still seems to work, but probably should do more testing once we have larger amounts deposited.


earn/src/data/BalanceSheet.ts
```diff
@@ -3,54 +3,29 @@ import Big from 'big.js';
...
-    reward += (shortfall * price) / ALOE_II_LIQUIDATION_INCENTIVE;
-  }
-  if (liabilities1 > assets1) {
-    const shortfall = liabilities1 - assets1;
-    reward += shortfall / ALOE_II_LIQUIDATION_INCENTIVE;
-  }
-  return reward;
+  return [
+    prices.a.gt(MIN_SQRT_RATIO) ? prices.a : MIN_SQRT_RATIO,
+    prices.b.lt(MAX_SQRT_RATIO) ? prices.b : MAX_SQRT_RATIO,
```
IanWoodard, line 27: We could use GN.max and GN.min here

prime/src/data/BalanceSheet.ts
```diff
@@ -3,53 +3,28 @@ import { GN } from 'shared/lib/data/GoodNumber';
...
+const PROBE_SQRT_SCALER_MAX = 3.078745359035;
 
-  sigma = sigma.recklessMul(nSigma);
-
-  let a = sqrtMeanPriceX96.mul(ONE.sub(sigma).sqrt());
-  let b = sqrtMeanPriceX96.mul(ONE.add(sigma).sqrt());
-
-  if (a.lt(MIN_SQRT_RATIO)) a = MIN_SQRT_RATIO.recklessAdd(1);
-  if (b.gt(MAX_SQRT_RATIO)) b = MAX_SQRT_RATIO.recklessSub(1);
+function _computeProbePrices(sqrtMeanPriceX96: GN, iv: GN, nSigma: number): [GN, GN] {
```
IanWoodard, line 15: Why are we still using different logic than in earn? Shouldn't we make them the same?

IanWoodard (approved):

earn/src/data/BalanceSheet.ts
haydenshively: These are Bigs, not GNs

prime/src/data/BalanceSheet.ts
haydenshively: I know it's kinda annoying, but I didn't think it was worth the time right now.

PR #645, opened 2023-09-24
Adding selected margin account as search param

IanWoodard:
Currently, the first margin account is selected whenever you refresh the page. With this change, it remembers previously selected ones when you refresh.


haydenshively (approved):
Sweet

PR #646, opened 2023-09-24
Update USDC; add IL to import widget and explanation to Boost page

haydenshively:


earn/src/components/boost/ImportBoostWidget.tsx
```diff
@@ -468,25 +497,36 @@ export default function ImportBoostWidget(props: ImportBoostWidgetProps) {
         </StyledDatalist>
       </SliderContainer>
       <Text size='M' color={SECONDARY_COLOR} className='mt-4'>
-        Estimated Fees
+        Estimated Earnings
       </Text>
       <div className='flex justify-center gap-2 mt-2'>
-        <div className='w-full'>
+        <div className='flex flex-row justify-center items-end'>
```
IanWoodard, line 503: why remove w-full?

IanWoodard (approved):

earn/src/components/boost/ImportBoostWidget.tsx
haydenshively: I don't remember, but it looks better

earn/src/pages/BoostPage.tsx
```diff
@@ -235,11 +265,17 @@ export default function BoostPage() {
...
+          Aloe Boost puts your Uniswap positions on steroids－earn up to 5x more swap fees and spend less on gas. Click
+          a lightning bolt to get started.
+        </Text>
+      </ExplainerWrapper>
+      <Text size='XL'>Positions</Text>
       <div className='flex flex-wrap gap-4 mt-4'>
         {isLoading &&
           allCardInfos.length === 0 &&
-          [...Array(4)].map((_, index) => <BoostCardPlaceholder key={index} />)}
+          [...Array(1)].map((_, index) => <BoostCardPlaceholder key={index} />)}
```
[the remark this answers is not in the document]
haydenshively: Well ideally the number of placeholders would match the number of cards, but that's not possible. So I thought it looked better to go from too few --> too many than vice versa. If we go from 4 placeholders to 2 actual cards, for example, it might make the user think that something is based.

PR #651, opened 2023-10-03
Adding last updated to stats card

IanWoodard:
Adding last updated time to stats card.


earn/src/pages/InfoPage.tsx
```diff
@@ -215,9 +216,29 @@ export default function InfoPage() {
...
       });
 
+      // Get the time at which each pool was last updated via the oracle (using the update event and getLogs)
+      const latestTimestamps = await Promise.all(
+        poolAddresses.map(async (addr) => {
+          let updateLogs: ethers.providers.Log[] = [];
+          try {
+            updateLogs = await provider.getLogs({
+              address: ALOE_II_ORACLE_ADDRESS[chainId],
+              topics: [TOPIC0_UPDATE_ORACLE, null, `0x000000000000000000000000${addr.slice(2)}`],
```
haydenshively, line 226: I think the pool address is the first indexed entry, not the second.

earn/src/pages/InfoPage.tsx
```diff
@@ -215,9 +216,29 @@ export default function InfoPage() {
...
+              fromBlock: 0,
+              toBlock: 'latest',
+            });
+            if (updateLogs.length > 0) {
+              const latestUpdate = updateLogs[updateLogs.length - 1].blockNumber;
```
haydenshively, line 231: Probably want to filter for logs where `removed === false`

See return value docs here: https://www.quicknode.com/docs/ethereum/eth_getLogs

earn/src/pages/InfoPage.tsx
```diff
@@ -215,9 +216,29 @@ export default function InfoPage() {
         }
       });
 
+      // Get the time at which each pool was last updated via the oracle (using the update event and getLogs)
+      const latestTimestamps = await Promise.all(
+        poolAddresses.map(async (addr) => {
```
haydenshively, line 221: Since you're iterating over all poolAddresses anyway, you could just _not_ filter logs by the address. The only downside with that is that if one of the pools hasn't been updated in a while, it may be so old as to not appear in the list.

Does alchemy return an unlimited number of logs? Or is this paginated somehow. If it's unlimited, then I'd definitely remove the filter and just fetch all the logs once --> filter locally.

earn/src/components/info/MarketCard.tsx
```diff
@@ -82,6 +94,10 @@ export default function MarketCard(props: MarketCardProps) {
 
   const etherscanLink = `${getEtherscanUrlForChain(activeChain)}/address/${poolAddress}`;
 
+  const lastUpdated = lastUpdatedAt
+    ? formatDistanceToNow(new Date(lastUpdatedAt * 1000), { includeSeconds: false, addSuffix: true })
+    : 'Never';
```
haydenshively, line 99: Naming?

haydenshively (commented):
Directionally correct. A few concerns on the fetching

earn/src/components/info/MarketCard.tsx
IanWoodard: What naming?

haydenshively: `lastUpdated` vs `lastUpdatedAt` is confusing.

earn/src/pages/InfoPage.tsx
```diff
@@ -217,18 +217,24 @@ export default function InfoPage() {
...
+          toBlock: 'latest',
+        });
+      } catch (e) {
+        console.error(e);
+      }
+      const filteredLogs = updateLogs.filter((log) => log.removed === false);
       const latestTimestamps = await Promise.all(
         poolAddresses.map(async (addr) => {
-          let updateLogs: ethers.providers.Log[] = [];
+          const associatedLogs = filteredLogs.filter((log) => log.topics[1] === `0x000000000000000000000000${addr}`);
```
haydenshively, line 234: Assuming incoming logs are sorted chronologically (an assumption you're already making on line 237), we don't need to do a full filter. We can instead just find the _first matching element_, starting from the end of the array and going backwards.

PR #680, opened 2023-10-15
Adding initial supply modal

IanWoodard:
This is subject to change and is merely a first iteration.
image

### DIFF
--- a/earn/src/components/lend/SupplyTable.tsx
+++ b/earn/src/components/lend/SupplyTable.tsx
@@ -8,6 +8,7 @@ import { formatTokenAmount } from 'shared/lib/util/Numbers';
 import styled from 'styled-components';
 
 import { TokenIconsWithTooltip } from '../common/TokenIconsWithTooltip';
+import SupplyModal from './modal/SupplyModal';
 import OptimizeButton from './OptimizeButton';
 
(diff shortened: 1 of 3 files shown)


earn/src/components/lend/SupplyTable.tsx
```diff
@@ -62,85 +64,101 @@ export default function SupplyTable(props: SupplyTableProps) {
...
-                  {formatTokenAmount(row.supplyBalance)} {row.asset.symbol}
-                </Display>
-              </td>
-              <td className='px-4 py-2 flex justify-center whitespace-nowrap'>
-                <OptimizeButton
-                  isOptimized={row.isOptimized}
-                  onClick={(e) => {
-                    e.stopPropagation();
-                  }}
+    <>
```
haydenshively, line 67: Why?

haydenshively (approved):
I'd left-align the collateral asset icons, otherwise looks good. Also note that if we finish the APY optimizer in time, the transaction logic will change.

earn/src/components/lend/SupplyTable.tsx
IanWoodard: So we can include the modal

PR #690, opened 2023-10-27
Allow users to reuse NFTs for Boost instead of minting new ones

haydenshively:
If users have an unused Borrower (e.g., they previously had a Boost position but burned it) in the appropriate pool, the import page will automatically use it instead of minting a new one. This saves gas and avoids state bloat (Like 800,000 --> 400,000 gas which is a lot).

Also created a more flexible BorrowerNft fetching function. It has various filtering options which I think will be sufficient for everything we want to do in the immediate future.


earn/src/data/BorrowerNft.ts
```diff
@@ -0,0 +1,122 @@
...
+  onlyCheckMostRecentModify: boolean;
+  includeFreshBorrowers: boolean;
+};
+
+export async function fetchListOfBorrowerNfts(
+  chainId: number,
+  provider: ethers.providers.BaseProvider,
+  userAddress: string,
+  filterParams?: BorrowerNftFilterParams
+) {
```
IanWoodard, line 24: Could we make the return type explicit?

earn/src/data/BorrowerNft.ts
```diff
@@ -0,0 +1,122 @@
...
+      const isInCorrectPool = filterParams?.validUniswapPool === undefined || borrowersInCorrectPool.has(borrower);
+      return (areManagersValid || (filterParams?.includeFreshBorrowers && !isInUse)) && isInCorrectPool;
+    })
+    .map(([borrower, managerSet]) => borrower);
+
+  // Fetch decoded SSTORE2 data from the BorrowerNFT (tokenIds in a specific order)
+  const orderedTokenIds = (await borrowerNftContract.tokensOf(userAddress)) as BigNumber[];
+  const orderedTokenIdStrs = orderedTokenIds.map((id) => id.toHexString());
+
+  const tokenIds = borrowers.map((borrower) => orderedTokenIdStrs.find((x) => x.startsWith(borrower.toLowerCase()))!);
```
IanWoodard, line 116: What happens when there is no x that starts with borrower.toLowerCase?

earn/src/data/BorrowerNft.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // Execute multicall fetch
+    const multicall = new Multicall({
+      ethersProvider: provider,
+      tryAggregate: true,
+      multicallCustomContractAddress: MULTICALL_ADDRESS[chainId],
+    });
+    const lensResults = (await multicall.call(lensContext)).results['lens'].callsReturnContext;
+    if (lensResults.find((v) => !v.success || !v.decoded)) {
+      throw new Error(`Multicall error while checking whether Borrowers are in use`);
```
IanWoodard, line 85: Why use ticks instead of single quotes?
```suggestion
      throw new Error('Multicall error while checking whether Borrowers are in use');
```

earn/src/components/boost/ImportBoostWidget.tsx
```diff
@@ -394,26 +401,68 @@ export default function ImportBoostWidget(props: ImportBoostWidgetProps) {
...
+    }
+    return marketParameters.ante;
+  }, [marketParameters, availableNft, borrowerBalance]);
+
   // Prepare for actual import/mint transaction
   const borrowerNft = useMemo(() => new ethers.utils.Interface(borrowerNftAbi), []);
   // First, we `mint` so that they have a `Borrower` to put stuff in
   const encodedMint = useMemo(() => {
-    if (!userAddress) return undefined;
+    if (!userAddress) return '0x';
```
IanWoodard, line 450: Why change?

earn/src/components/boost/ImportBoostWidget.tsx
```diff
@@ -394,26 +401,68 @@ export default function ImportBoostWidget(props: ImportBoostWidgetProps) {
...
     const to = userAddress;
     const pools = [cardInfo.uniswapPool];
     const salts = [generateBytes12Salt()];
     return borrowerNft.encodeFunctionData('mint', [to, pools, salts]) as `0x${string}`;
   }, [borrowerNft, userAddress, cardInfo]);
   // Then we `modify`, calling the BoostManager to import the Uniswap position
   const encodedModify = useMemo(() => {
-    if (!userAddress || nextNftPointerIndex === undefined) return undefined;
+    if (!userAddress || nextNftPtrIdx === undefined) return '0x';
```
IanWoodard, line 458: Why change?

IanWoodard (approved):

earn/src/data/BorrowerNft.ts
haydenshively: There must be such an `x` based on the contract logic, unless something has gone horribly wrong.

earn/src/components/boost/ImportBoostWidget.tsx
haydenshively: To get rid of the annoying ternary in the usePrepareContractWrite call. Because now we have _another_ ternary there, so if we left this as returning undefined it would be nested ternaries.

earn/src/components/boost/ImportBoostWidget.tsx
haydenshively: See above

earn/src/data/BorrowerNft.ts
haydenshively: I don't really love this return type, though its functional for now. I don't really want to make it explicit because then we're less likely to come up with a better one in the future.

earn/src/data/BorrowerNft.ts
haydenshively: Leftover from the place I copy/pasted from (we were formatting with a Borrower address inside or something)

PR #691, opened 2023-10-27
Temporarily remove optimize button

IanWoodard:
image


haydenshively (approved):
Thanks. We should probably have a maximum width on this table now so it doesn't look weird.

PR #700, opened 2023-11-03
Adding deposit logic to markets page

IanWoodard:
Re-using most of the logic from the `EarnInterestModal`.


haydenshively (approved):
LGTM

PR #704, opened 2023-11-05
Adding separate loading state for boostCardInfos

IanWoodard:


haydenshively:
Wait why is this necessary?

IanWoodard:
> Wait why is this necessary?

Because there are two main things being fetched, and if the uniswapNFT positions load first, it will look like there are no boostedCards even if there are.

PR #709, opened 2023-11-10
Tweaking markets page

IanWoodard:
Makes the entire tab button clickable, makes it more accessible, and adjusts the colors.
image


earn/src/pages/MarketsPage.tsx
```diff
@@ -30,17 +30,17 @@ import { fetchBorrowerDatas, MarginAccount } from '../data/MarginAccount';
 import { PriceRelayLatestResponse } from '../data/PriceRelayResponse';
 import { getProminentColor } from '../util/Colors';
 
+const UNSELECTED_OPTION_TEXT_COLOR = 'rgba(75, 105, 128, 1)';
```
haydenshively, line 33: Where is this coming from? If it's from the spec, we should define it in shared, and if not, let's switch to something that is -- will grey 600 or 700 work?

earn/src/pages/MarketsPage.tsx
```diff
@@ -52,7 +52,7 @@ const HeaderSegmentedControlOption = styled(Text).attrs((props: { isActive: bool
     content: ${(props) => (props.isActive ? "''" : 'none')};
     background-color: white;
 
-    top: 40px;
+    top: 100%;
```
haydenshively, line 55: nifty

earn/src/pages/MarketsPage.tsx
```diff
@@ -358,23 +358,27 @@ export default function MarketsPage() {
...
-          </HeaderSegmentedControlOption>
-          <HeaderSegmentedControlOption
-            size='M'
-            isActive={selectedHeaderOption === HeaderOptions.Borrow}
-            onClick={() => setSelectedHeaderOption(HeaderOptions.Borrow)}
-          >
-            Borrow
-          </HeaderSegmentedControlOption>
+        <div>
+          <div className='flex flex-row' role='tablist'>
```
haydenshively, line 362: 👍

haydenshively (approved):

earn/src/pages/MarketsPage.tsx
IanWoodard: It is the same color used for the navbar

IanWoodard: Just checked, this is GREY_400. Will add it to shared

PR #783, opened 2024-01-16
Fixing perpetual leaderboard queries

IanWoodard:
Currently, the leaderboard is queried repeatedly in an infinite loop.


haydenshively (approved):
Oops my b

PR #786, opened 2024-01-24
Refactor handling of LendingPairBalances (prep for BorrowWidget updates)

haydenshively:
When the user clicks on an available borrow asset (right side of the `BorrowingWidget`), I noticed that the collateral entries still show a range of LTVs. Similar to how APYs are narrowed down once you make a selection on the left side, LTVs should be narrowed down once you make a selection on the right side.

With the way things were, this would be very hard to do -- the collateral entries were being pre-computed _outside_ of the `BorrowingWidget` component, and to be honest the multitude of arrays (`LendingPairBalances[]`, `combinedBalances`, `tokenBalances`, etc.) were hard to reason about.

So, to prepare for the `BorrowingWidget` updates I want to make, I refactored how we fetch `LendingPairBalances`. It's fully backwards-compatible with existing logic (so I didn't change this stuff on `PortfolioPage` and `LendPage`), but it allows us to use a map instead of searching through an array over and over.

The net result is less code, fewer renders, and fewer network requests (~130 --> ~100). Also, each network request is smaller because I'm making fewer duplicate contract calls in each `multicall`.

### DIFF
--- a/earn/src/components/portfolio/modal/SendCryptoModal.tsx
+++ b/earn/src/components/portfolio/modal/SendCryptoModal.tsx
@@ -32,7 +32,7 @@ function getConfirmButton(state: ConfirmButtonState, token: Token): { text: stri
   switch (state) {
     case ConfirmButtonState.INVALID_ADDRESS:
       return {
-        text: `Invalid Address`,
+        text: `Couldn't resolve ENS`,
         enabled: false,
       };
     case ConfirmButtonState.INSUFFICIENT_ASSET:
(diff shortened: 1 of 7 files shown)


earn/src/components/portfolio/modal/SendCryptoModal.tsx
```diff
@@ -32,7 +32,7 @@ function getConfirmButton(state: ConfirmButtonState, token: Token): { text: stri
   switch (state) {
     case ConfirmButtonState.INVALID_ADDRESS:
       return {
-        text: `Invalid Address`,
+        text: `Couldn't resolve ENS`,
```
haydenshively, line 35: Unrelated change; I'm aware; it's small enough it doesn't matter imo

IanWoodard (approved):

PR #825, opened 2024-02-24
Handle no lens results

IanWoodard:
Adding logic to handle and return early the case where there are no lens results when fetching a list of borrower nfts.

### DIFF
--- a/earn/src/data/BorrowerNft.ts
+++ b/earn/src/data/BorrowerNft.ts
@@ -106,7 +106,9 @@ export async function fetchListOfBorrowerNfts(
       tryAggregate: true,
       multicallCustomContractAddress: MULTICALL_ADDRESS[chainId],
     });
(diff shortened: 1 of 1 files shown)


haydenshively (commented):
We should investigate why the lens call is failing. That shouldn't happen.

IanWoodard:
> We should investigate why the lens call is failing. That shouldn't happen.

While I agree, it is failing, so we should have logic to handle such a case.

haydenshively:
Turns out this happens because the multicall context has zero calls. To fix, we should change the enclosing if statement to look like this:

```ts
  // Fetch borrowers' in-use status and Uniswap pool, which we'll need for filtering
  if (
    (filterParams?.includeFreshBorrowers || Boolean(filterParams?.validUniswapPool)) &&
    borrowerManagerSets.size > 0
  )
```

(testing that the set is non-empty is the key)

PR #829, opened 2024-02-25
Filter borrowers to prevent blank positions

IanWoodard:
Prevents the active borrowers section from showing up when it shouldn't

### DIFF
--- a/earn/src/components/markets/borrow/BorrowingWidget.tsx
+++ b/earn/src/components/markets/borrow/BorrowingWidget.tsx
@@ -272,7 +272,7 @@ export default function BorrowingWidget(props: BorrowingWidgetProps) {
 
   return (
     <>
-      {(filteredBorrowers?.length || 0) > 0 && (
+      {Boolean(filteredBorrowers?.length) && (
         <>
           <Text size='L' weight='bold'>
             Manage positions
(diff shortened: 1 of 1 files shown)


earn/src/components/markets/borrow/BorrowingWidget.tsx
```diff
@@ -277,72 +287,63 @@ export default function BorrowingWidget(props: BorrowingWidgetProps) {
...
+                    })}
                   </div>
                 </CardRow>
               </CardContainer>
             </CardWrapper>
             <div className='w-[52px] mt-[2px]'>
               <div className='w-[52px] h-[42px]' />
-              {borrowers &&
-                borrowers.map((borrower) => {
-                  const hasNoCollateral = borrower.assets.amount0.isZero() && borrower.assets.amount1.isZero();
```
haydenshively, line 338: Functional difference here in that we will now check that Uniswap positions are non-zero. Is that good?

haydenshively (approved):

earn/src/components/markets/borrow/BorrowingWidget.tsx
IanWoodard: I couldn't think of a good reason we had it the way we did. If you can think of one, lmk. My thought was that we wouldn't want to show the health if there is no borrower to show.

PR #836, opened 2024-03-04
Adding selectedTab search param

IanWoodard:
This PR enables users to remain on the same tab on the markets page when they refresh.


haydenshively (approved):
🦾

PR #838, opened 2024-03-09
Fixing a bug that causes the markets page to crash

IanWoodard:
Fixing a bad state that results from switching chains when on a page besides the first on the markets page supply tab.


haydenshively (approved):
This might happen on the Portfolio page too

IanWoodard:
> This might happen on the Portfolio page too

I don't think we have any concept of pages on the portfolio page, but I can take a look.

PR #841, opened 2024-03-23
Prevent overflow on portfolio page

IanWoodard:
Added some additional styling to prevent overflowing on the lending pair peer card. In addition to adding this, we should probably also investigate a better technique for figuring out how many sigdigs to use.
Before:
image
After:
image

### DIFF
--- a/earn/src/components/portfolio/LendingPairPeerCard.tsx
+++ b/earn/src/components/portfolio/LendingPairPeerCard.tsx
@@ -79,7 +79,7 @@ const CardHeader = styled.div`
 
 const CardBody = styled.div`
   display: grid;
-  grid-template-columns: 3fr 1fr 1fr 1fr;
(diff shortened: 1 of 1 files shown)


haydenshively (commented):
This should probably be scientific notation. Or for Ether specifically, we could use units like gwei and wei.

IanWoodard:
> This should probably be scientific notation. Or for Ether specifically, we could use units like gwei and wei.

Read the rest of my description, I explain that

IanWoodard:
> > This should probably be scientific notation. Or for Ether specifically, we could use units like gwei and wei.
> 
> Read the rest of my description, I explain that

This PR just ensures it doesn't overflow, it's a failsafe

haydenshively (approved):
Understood. Originally interpreted it as "we will replace this with better formatting" when you really meant "this is a failsafe if our better formatting still sucks"

LGTM

PR #843, opened 2024-03-30
Adding slippage for importing boost positions

IanWoodard:
The slippage can be adjusted by clicking on the cog which shows the following menu:
image

### DIFF
--- a/earn/src/components/boost/ImportBoostWidget.tsx
+++ b/earn/src/components/boost/ImportBoostWidget.tsx
@@ -45,7 +45,7 @@ import { BoostCardInfo } from '../../data/Uniboost';
 import { getValueOfLiquidity, UniswapPosition, UniswapV3GraphQL24HourPoolDataQueryResponse } from '../../data/Uniswap';
 import { BOOST_MAX, BOOST_MIN } from '../../pages/boost/ImportBoostPage';
 import { getTheGraphClient, Uniswap24HourPoolDataQuery } from '../../util/GraphQL';
-import MaxSlippageInput from '../common/MaxSlippageInput';
+import SlippageWidget from './SlippageWidget';
 
 const SECONDARY_COLOR = '#CCDFED';
 const TERTIARY_COLOR = '#4b6980';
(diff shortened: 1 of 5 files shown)


earn/src/components/boost/ImportBoostWidget.tsx
```diff
@@ -275,6 +277,16 @@ export default function ImportBoostWidget(props: ImportBoostWidgetProps) {
...
+    const maxSlippage = parseFloat(maxSlippagePercentage) / 100;
+    const maxBorrowAmount0 = borrowAmount0
+      .recklessMul(1 + maxSlippage)
+      .toBigNumber()
+      .add(2);
+    const maxBorrowAmount1 = borrowAmount1
+      .recklessMul(1 + maxSlippage)
+      .toBigNumber()
+      .add(2);
+    const combinedMaxBorrowAmount = maxBorrowAmount0.add(maxBorrowAmount1.mul(GN.Q(112).toBigNumber()));
```
haydenshively, line 289: If `BigNumber` has a `shl` function, I would shift left 112 instead of converting Q112 to a BigNumber and multiplying.

haydenshively: Looks like `shiftedBy` is a thing

shared/src/util/Numbers.ts
```diff
@@ -135,7 +135,9 @@ export function roundUpToNearestN(value: number, n: number): number {
 
 export function formatTokenAmount(amount: number, sigDigs = 4): string {
   //TODO: if we want to support more than one locale, we would need to add more logic here
-  if (amount > 1e13) {
+  if (amount > 1e34) {
+    return '∞';
```
haydenshively, line 139: Hmm I'm not sure I like this. Can we make this more specific to the problem you're trying to solve?

haydenshively (approved):

earn/src/components/boost/ImportBoostWidget.tsx
IanWoodard: Wrong BigNumber library, but ethers Bignumber does have `shl`.

PR #844, opened 2024-03-30
Adding token allocation to the advanced page

IanWoodard:
This PR adds a token allocation section to the advanced page to show the relative breakdown of the raw tokens vs Uniswap positions.
Wider screens:
image
Narrower screens:
image
Empty state:
image

### DIFF
--- a/earn/src/components/advanced/TokenAllocationPieChartWidget.tsx
+++ b/earn/src/components/advanced/TokenAllocationPieChartWidget.tsx
@@ -16,7 +16,7 @@ import { rgb, rgba } from '../../util/Colors';
 const PIE_CHART_HOVER_GROWTH = 1.05;
 
 const PieChartContainer = styled.div`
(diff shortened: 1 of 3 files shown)


IanWoodard:
> What do you think of this? image

I prefer them to be stacked vertically since it makes it easier to see when hovering over the pie chart.

earn/src/components/advanced/TokenAllocationPieChartWidget.tsx
```diff
@@ -0,0 +1,414 @@
...
+  ${tw`flex flex-col justify-center items-center`};
+  & div {
+    transition: color 0.15s linear;
+  }
+`;
+
+const TokenLabel = styled.div`
+  width: 90px;
+  font-size: 20px;
+  font-weight: 300;
```
IanWoodard, line 116: Not a fan of how thin the text looks here

PR #848, opened 2024-04-12
Allow users to deposit raw ETH instead of just WETH

haydenshively:


earn/src/components/portfolio/modal/EarnInterestModal.tsx
```diff
@@ -285,8 +307,8 @@ export default function EarnInterestModal(props: EarnInterestModalProps) {
             <BaseMaxButton
               size='L'
               onClick={() => {
-                if (depositBalance != null) {
-                  setInputValue(depositBalance?.formatted);
+                if (tokenBalanceResult != null) {
+                  setInputValue(userBalance.toString(GNFormat.DECIMAL));
```
IanWoodard, line 311: While I see why it was done this way, I wish we could avoid having seemingly unrelated variables used in the if statement (despite them being quite related in actuality).

earn/src/data/LendingPair.ts
```diff
@@ -360,6 +362,12 @@ export async function getLendingPairBalances(
   const deprecatedLendingPairBalancesArray: LendingPairBalances[] = [];
   const balancesMap: LendingPairBalancesMap = new Map();
 
+  balancesMap.set('0x0000000000000000000000000000000000000000', {
```
IanWoodard, line 365: Could we make this (empty address) a constant to avoid silly mistakes?

IanWoodard (approved):
Left some comments, please make sure to check prime works too

earn/src/components/portfolio/modal/EarnInterestModal.tsx
```diff
@@ -215,18 +223,27 @@ export default function EarnInterestModal(props: EarnInterestModalProps) {
...
   // Get the user's balance of the selected token
-  const { refetch: refetchDepositBalance, data: depositBalance } = useBalance({
+  const { refetch: refetchBalanceToken, data: tokenBalanceResult } = useBalance({
     address: account?.address ?? '0x',
     token: selectedOption.address,
     chainId: activeChain.id,
     enabled: isOpen,
   });
 
+  const isWeth = selectedOption.name === 'Wrapped Ether';
```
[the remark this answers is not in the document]
haydenshively: No they're not, and I believe I put a TODO somewhere about finding a better way to do this.

That said, we're in control of our own token list, so as long as _we_ don't add multiple tokens on a single chain called 'Wrapped Ether', we're fine.

shared/src/data/constants/ChainSpecific.tsx
```diff
@@ -42,6 +43,13 @@ export const APPROX_SECONDS_PER_BLOCK: { [chainId: number]: number } = {
   [base.id]: 2,
 };
 
+export const ETH_RESERVED_FOR_GAS: { [chainId: number]: GN } = {
+  [mainnet.id]: GN.fromDecimalString('0.1', 18),
+  [optimism.id]: GN.fromDecimalString('0.005', 18),
+  [arbitrum.id]: GN.fromDecimalString('0.005', 18),
+  [base.id]: GN.fromDecimalString('0.005', 18),
+};
```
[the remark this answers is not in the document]
haydenshively: image

earn/src/pages/PortfolioPage.tsx
```diff
@@ -84,12 +84,12 @@ export type PriceEntry = {
 };
 
 export type TokenQuote = {
-  token: Token;
+  token?: Token;
   price: number;
```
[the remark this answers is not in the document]
haydenshively: In general, yes, obviously. However, when the source of symbols (price relay) doesn't map onto the local list (`TokenData`), it makes sense.

This is the change that allowed my branch to handle the price relay reporting "ETH" without an "ETH" entry in `TokenData`. To mitigate future mistakes of the same sort, I think we should keep it like this.

earn/src/components/portfolio/modal/EarnInterestModal.tsx
haydenshively: Hmm good point. This was actually just a result of refactoring; I didn't write the code, but now that you point it out it is kinda bad/confusing. Any suggestions to fix?

earn/src/data/LendingPair.ts
haydenshively: Turns out we already had one :)

PR #850, opened 2024-04-16
Make borrowed funds dst and repayment src configurable on Advanced page

haydenshively:
Standard procedure is to immediately send borrowed funds to the user's wallet, and pull from the same in order to repay.

This change allows for an alternate path, where the user borrows but _leaves funds in the NFT_, or (similarly) use funds in the NFT to repay, instead of using funds from their wallet.

This can eliminate a few extra txns depending on the use-case, and will be particularly useful once we allow people to create Uniswap positions through the Advanced page

### DIFF
--- a/earn/src/components/advanced/modal/BorrowModal.tsx
+++ b/earn/src/components/advanced/modal/BorrowModal.tsx
@@ -1,5 +1,6 @@
 import { useContext, useState, useMemo, useEffect } from 'react';
 
+import { CheckIcon } from '@heroicons/react/solid';
 import { Address, SendTransactionResult } from '@wagmi/core';
 import { ethers } from 'ethers';
 import { borrowerAbi } from 'shared/lib/abis/Borrower';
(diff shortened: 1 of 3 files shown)


earn/src/components/advanced/modal/BorrowModal.tsx
```diff
@@ -232,6 +244,7 @@ export default function BorrowModal(props: BorrowModalProps) {
   }
 
   // TODO: use GN
+  // TODO: if !shouldWithdrawToWallet, max borrows should just be from `maxBorrow`, not `maxBorrowAndWithdraw`
```
IanWoodard, line 247: Should we implement this in this PR?

earn/src/components/advanced/modal/RepayModal.tsx
```diff
@@ -192,19 +207,142 @@ function RepayButton(props: RepayButtonProps) {
...
+      userAddress,
+      [borrower.index],
+      [ALOE_II_BORROWER_NFT_SIMPLE_MANAGER_ADDRESS[activeChain.id]],
+      [encodedData as `0x${string}`],
+      [0],
+    ],
+    enabled: Boolean(userAddress) && repayAmount.isGtZero(),
+    chainId: activeChain.id,
+  });
+  // NOTE: Not using `useMemo` to update the request
```
IanWoodard, line 265: Remove?

IanWoodard (approved):

earn/src/components/advanced/modal/BorrowModal.tsx
haydenshively: We should probably do it soon, but it's not absolutely necessary. The new one would be strictly greater than the current one, so it's not like it'll cause the txn to fail.

And to be honest I don't feel like doing that math right now.

PR #853, opened 2024-04-17
Add oracle cardinality to Monitor tab

haydenshively:
Our oracle needs 60 minutes of data to work right. We require 3 hours of data at pool creation time, but it's possible to stuff the oracle with extra data via wash trading and _shorten the lookback window_. Therefore, I think it's important to surface the actual oracle cardinality on the frontend so users can make even more informed decisions about where to deposit.

### DIFF
--- a/earn/src/components/markets/monitor/StatsTable.tsx
+++ b/earn/src/components/markets/monitor/StatsTable.tsx
@@ -217,6 +217,9 @@ function StatsTableRow(props: StatsTableRowProps) {
         <Display size='XS'>{reserveFactorText}</Display>
       </td>
       <td className='px-4 py-2 text-start whitespace-nowrap'>
+        <Display size='XS'>{pair.slot0Data.observationCardinality}</Display>
+      </td>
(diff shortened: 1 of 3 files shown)


earn/src/components/markets/monitor/StatsTable.tsx
```diff
@@ -320,6 +323,11 @@ export default function StatsTable(props: { rows: StatsTableRowProps[] }) {
                   Reserve Factor
                 </Text>
               </th>
+              <th className='px-4 py-2 text-start whitespace-nowrap'>
+                <Text size='M' weight='bold'>
+                  Cardinality
+                </Text>
+              </th>
```
IanWoodard, line 330: Could we provide a description of what this is somewhere?

IanWoodard (approved):

earn/src/components/markets/monitor/StatsTable.tsx
haydenshively: Yes, I meant to ask you about adding some tooltips to a few of the columns.

Can you set up a component for that, and I can provide explanations?

PR #877, opened 2024-06-22
Linting

IanWoodard:
I ran the linter locally and it updated these lines.


haydenshively (approved):
Weird. Relatively few changes though so go ahead and merge.

PR #878, opened 2024-06-22
Fixing tests

IanWoodard:
After upgrading to react 18, some of our tests were showing warnings, this change should clean those up.


haydenshively (approved):
💪

